PR #21, opened 2019-04-22
Update to graphql-java v12 + required modifications

lilianaziolek:
- upgrade to latest graphql version -> added generics on Type
- improve implementation of schema augmentation - lists are now immutable so have to be recreated and reset
- directives are added to schema as schema validation is now much stricter
- preventing _id duplicates in query params
- declare relation direction as enum
- fixes on tests


lilianaziolek:
Closing, will reopen from separate branches to split changes into multiple PRs

PR #23, opened 2019-04-25
Update to graphql-java v12

lilianaziolek:
* upgrade to latest graphql version -> added generics on Type
* improve implementation of schema augmentation - lists are now immutable so have to be recreated and reset
* directives are added to schema as schema validation is now much stricter
* preventing _id duplicates in query params
* declare relation direction as enum
* fixes on tests
(Note this is reopen PR as the previous was raised from master not feature branch, thus picking up too many changes)

### DIFF
--- a/src/main/kotlin/org/neo4j/graphql/Translator.kt
+++ b/src/main/kotlin/org/neo4j/graphql/Translator.kt
@@ -1,17 +1,17 @@
 package org.neo4j.graphql
 
 import graphql.Scalars
-import graphql.introspection.Introspection
 import graphql.language.*
 import graphql.parser.Parser
-import graphql.schema.*
+import graphql.schema.GraphQLFieldDefinition
+import graphql.schema.GraphQLObjectType
+import graphql.schema.GraphQLSchema
+import graphql.schema.GraphQLType
 import graphql.schema.idl.RuntimeWiring
 import graphql.schema.idl.SchemaGenerator
 import graphql.schema.idl.SchemaParser
 import graphql.schema.idl.TypeDefinitionRegistry
 import org.antlr.v4.runtime.misc.ParseCancellationException
-import java.lang.RuntimeException
-import java.util.*
 
 class Translator(val schema: GraphQLSchema) {
     data class Context @JvmOverloads constructor(val topLevelWhere: Boolean = true,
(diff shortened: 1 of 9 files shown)


src/main/kotlin/org/neo4j/graphql/Translator.kt
```diff
@@ -359,75 +359,79 @@ class Translator(val schema: GraphQLSchema) {
...
-                schemaDefinition.operationTypeDefinitions.add(OperationTypeDefinition("query", TypeName(it.name)))
-                typeDefinitionRegistry.add(it)
-            }
-        }
-        val mutationDefinition = operations.getOrElse("mutation") {
-                ObjectTypeDefinition("Mutation").also {
-                    schemaDefinition.operationTypeDefinitions.add(OperationTypeDefinition("mutation", TypeName(it.name)))
-                    typeDefinitionRegistry.add(it)
+            typeDefinitionRegistry.getType("Query", ObjectTypeDefinition::class.java).orElseGet {
+                ObjectTypeDefinition("Query").also { typeDefinitionRegistry.add(it) }
```
jexp: it's not fully equivalent to the before, as before we did add it if it was not there?

src/main/kotlin/org/neo4j/graphql/Translator.kt
```diff
@@ -359,75 +359,79 @@ class Translator(val schema: GraphQLSchema) {
...
+        """
+        typeDefinitionRegistry.merge(schemaParser.parse(directivesSdl))
 
         val augmentations = typeDefinitionRegistry.types().values.filterIsInstance<ObjectTypeDefinition>().map { augmentedSchema(ctx, it) }
 
         val augmentedTypesSdl = augmentations.flatMap { listOf(it.filterType, it.ordering, it.inputType).filter { it.isNotBlank() } }.joinToString("\n")
         typeDefinitionRegistry.merge(schemaParser.parse(augmentedTypesSdl))
 
-        val schemaDefinition = typeDefinitionRegistry.schemaDefinition().orElseGet { SchemaDefinition() }
-        if (!typeDefinitionRegistry.schemaDefinition().isPresent) typeDefinitionRegistry.add(schemaDefinition).ifPresent { throw RuntimeException(it.toSpecification().toString()) }
```
jexp: it's not fully equivalent to the before, as before we did add the schema if it was not there?

src/main/kotlin/org/neo4j/graphql/Translator.kt
```diff
@@ -359,75 +359,79 @@ class Translator(val schema: GraphQLSchema) {
...
-                .filter { it.query.isNotBlank() && queryDefinition.fieldDefinitions.none { fd -> it.query.startsWith(fd.name+"(") } }
+                .filter { it.query.isNotBlank() && queryDefinition.fieldDefinitions.none { fd -> it.query.startsWith(fd.name + "(") } }
                 .map { it.query }.let {
-            if (!it.isEmpty()) {
-                val newQueries = schemaParser.parse("type AugmentedQuery { ${it.joinToString("\n")} }").getType("AugmentedQuery").get() as ObjectTypeDefinition
-                queryDefinition.fieldDefinitions.addAll(newQueries.fieldDefinitions)
+                    if (!it.isEmpty()) {
+                        val newQueries = schemaParser.parse("type AugmentedQuery { ${it.joinToString("\n")} }").getType("AugmentedQuery").get() as ObjectTypeDefinition
+                        typeDefinitionRegistry.remove(queryDefinition)
+                        typeDefinitionRegistry.add(queryDefinition.transform { qdb -> newQueries.fieldDefinitions.forEach { qdb.fieldDefinition(it) } })
```
jexp: is this merging old and new query definitions?

jexp:
Looks great, thanks a lot!

jexp:
The enum fields should also be able to take strings, at least the spec was lenient around that in the past.

lilianaziolek:
For future generations ;) looking at the comments / question re: adding query/schema etc. - graphql objects are now effectively immutable as they return mutable copies of collections like operationTypeDefintions or directives. This means that the whole object needs to be set in full - you cannot just modify schema defintion on existing one. That's the reason for waiting to set it on typeDefinitionRegistry until we have full view of what it should be.

Issue #22, opened 2019-04-24
Support for relationships on the same node types


lilianaziolek:
When specifying a relation with 2 different node types on both ends (e.g. Movie and Actor), this relation can be used inside Movie or Actor and gets correct direction (depending on from/to values in relation).
If both ends of a relation have the same type, there currently seems to be no way to indicate which "side" of the relation a field should relate to. See an example below:
```
type User {
  name:String
  referredBy: Referral @relation(direction: OUT)
  referred:[Referral] @relation(direction: IN)
}
type Referral @relation (name:"REFERRED_BY", from:"user", to: "referredBy" ) {
  user:User
  referredBy:User
  referralDate:String
}
```
At the moment, generated query will use outgoing relationship for both referredBy and referred

PR #24, opened 2019-04-25
Bugfix for issue#22 - relation with same types does not generate correct cypher

lilianaziolek:
Note at this point it is a partial fix for issue #22 + corresponding tests (failing)
The original issue (direction of cypher relation) has now been fixed and translator takes into account direction of relation where both types on the relation are the same.

However, this fix has uncovered another issue - the generated cypher does not actually work and that is due to the fact that two different names of node are used. For example, a generated query looks as follows:
```
MATCH (user:User) WHERE user.name = $userName 
RETURN user { .name,
referredBy:[(user)-[userReferredBy:REFERRED_BY]->(userReferredByReferredBy:User) | userReferredBy { .referralDate,referredBy:userReferredByReferredBy { .name } }][0],
referred:[(user)<-[userReferred:REFERRED_BY]-(userReferredUser:User) | userReferred { .referralDate,user:userReferredReferredBy { .name } }] } AS user
```
Note that user:userReferredReferredBy is incorrect as userReferredReferredBy is undefined. I will update this PR if I managed to fix this issue.

Also note that this PR depends on PR#23 which upgrades graphql to V12 and makes direction property an enum

### DIFF
--- a/src/main/kotlin/org/neo4j/graphql/Translator.kt
+++ b/src/main/kotlin/org/neo4j/graphql/Translator.kt
@@ -1,17 +1,17 @@
 package org.neo4j.graphql
 
 import graphql.Scalars
-import graphql.introspection.Introspection
 import graphql.language.*
 import graphql.parser.Parser
-import graphql.schema.*
+import graphql.schema.GraphQLFieldDefinition
+import graphql.schema.GraphQLObjectType
+import graphql.schema.GraphQLSchema
+import graphql.schema.GraphQLType
 import graphql.schema.idl.RuntimeWiring
 import graphql.schema.idl.SchemaGenerator
 import graphql.schema.idl.SchemaParser
 import graphql.schema.idl.TypeDefinitionRegistry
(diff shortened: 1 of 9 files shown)


src/main/kotlin/org/neo4j/graphql/Translator.kt
```diff
@@ -271,7 +271,11 @@ class Translator(val schema: GraphQLSchema) {
                 ?: throw IllegalStateException("Field $field needs an @relation directive")
 
         var relInfo = relDetails(fieldObjectType, relDirective)
-        val inverse = isRelFromType && fieldObjectType.getFieldDefinition(relInfo.startField).type.inner().name != parent.name
+        val fieldRelationDirection = fieldDefinition.getDirective("relation")?.getArgument("direction")?.value?:"OUT"
```
jexp: this data should all be part of relInfo ... coming from relDetails ...

jexp: also we should encapsulate it in a function to make it more readable

PR #31, opened 2019-05-06
fixes #19: add merge operation

### LINKED ISSUE #19: add merge operation
to CrudOperations
to Translator.kt: line 75
and AugmentationTest

only if the type has an ID to merge on.

conker84:
Fixes #19 

Add merge operation only if the type has an ID to merge on.

### DIFF
--- a/src/main/kotlin/org/neo4j/graphql/Translator.kt
+++ b/src/main/kotlin/org/neo4j/graphql/Translator.kt
@@ -41,7 +41,7 @@ class Translator(val schema: GraphQLSchema) {
         val queryType = schema.queryType.fieldDefinitions.filter { it.name == name }.firstOrNull()
         val mutationType = schema.mutationType.fieldDefinitions.filter { it.name == name }.firstOrNull()
         val fieldDefinition = queryType ?: mutationType
-            ?: throw IllegalArgumentException("Unknown Query $name available queries: " + schema.queryType.fieldDefinitions.map { it.name }.joinToString())
+            ?: throw IllegalArgumentException("Unknown Query $name available queries: " + (schema.queryType.fieldDefinitions + schema.mutationType.fieldDefinitions).map { it.name }.joinToString())
         val isQuery = queryType != null
         val returnType = fieldDefinition.type.inner()
 //        println(returnType)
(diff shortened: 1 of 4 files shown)


src/test/resources/movie-test.md
```diff
@@ -339,3 +339,23 @@ RETURN movie { .title,actors:[(movie)<-[:ACTED_IN]-(movieActors:Actor) |
...
+  }
+}
+```
+
+```params
+{"actorUserId": "1", "actorName": "Andrea"}
+```
+
+```cypher
+MERGE (actor:Actor {userId:$actorUserId}) ON CREATE SET actor.userId=$actorUserId,actor.name=$actorName
```
jexp: we should use SET not 
ON CREATE SET

**and not set the id again**

src/test/resources/movies-test-schema.graphql
```diff
@@ -1,5 +1,5 @@
 type Movie {
-  _id: String
+#  _id: String
```
jexp: can you add an issue for the _id field ?

src/test/resources/movies-test-schema.graphql
```diff
@@ -75,14 +75,18 @@ type Book {
...
   MovieById(movieId: ID!): Movie
-  MovieBy_Id(_id: String!): Movie
   GenresBySubstring(substring: String): [Genre] @cypher(statement: "MATCH (g:Genre) WHERE toLower(g.name) CONTAINS toLower($substring) RETURN g")
   Books: [Book]
 }
 type Mutation {
   createGenre(name:String): Genre @cypher(statement:"CREATE (g:Genre) SET g.name = name RETURN g")
 }
 # scalar DateTime
+
```
jexp: can you also add an issue for this one?

src/main/kotlin/org/neo4j/graphql/Translator.kt
```diff
@@ -66,9 +66,15 @@ class Translator(val schema: GraphQLSchema) {
                 // todo extract method or better object
                 val properties = properties(variable, fieldDefinition, propertyArguments(field))
```
jexp: we can separate out all-properties and properties without-id which we can use later on in SET, UPDATE etc.

conker84: Is it good for you if we manage that in the `setProperties` method by providing a list of properties to exclude?

src/test/resources/movie-test.md
conker84: ok

src/test/resources/movies-test-schema.graphql
conker84: I restored the old version and with the actual changes all works, as well as for the _id field so we don't need to open an issue for them

conker84:
Thanks a lot for the feedback!
I updated the issue

src/main/kotlin/org/neo4j/graphql/Translator.kt
jexp: that's fine to, but then we should rename the method :)

jexp:
Thanks a lot.

PR #41, opened 2019-06-19
fixes #10: support date/time

### LINKED ISSUE #10: support date/time
support the Neo4j 3.4.x datetime (and localdatetime) values

- as inputs and outputs (todo how to render the output)
- research other approaches in GraphQL libraries for datetime
- see implementation in javascript library

conker84:
List of provided changes:
- Renamed `TemporalTypes` into `Neo4jTypes` so we can easily add the support for Point values and reuse the same logic
- Implemented the support to the Date types for queries and mutations


jexp:
Thanks a lot.

Issue #71, opened 2019-11-07
Invalid parameter names for range queries


DirkMahler:
Given the following type...

```
type Person {
  name: String!
  age: int!
}
```

... I want to execute a range query:

```
Person(filter: {
  age_gte: 20
  age_lt: 30
}) {
  name
}
```
The query currently returns an empty result, as the given parameters in the filter are collapsed into one `filterPersonAge`. The solution would be to use the full filter attribute name, e.g. ` filterPersonAge_gte` and `filterPersonAge_lt`.

jexp:
Great find, thanks Dirk

@Andy2003 could you have a look at that please?

DirkMahler:
Already working on it, I'll provide a PR for this.

PR #72, opened 2019-11-13
Append operand name in filter expressions

DirkMahler:
Solves #71


jexp:
Cool, thanks a lot Dirk !

Issue #63, opened 2019-09-25
performance issues of deep filters


jexp:
deep filters don't use the indexes but do label scan + filter

i.e. currently we generate `ALL(expr IN [(start)-->(next:Next) | next.name = $foo] WHERE expr)`

and nested variants of this for each level of nested filters

* we also do `NONE/SINGLE/SOME(expr ...)`

which we had to generate because:

1. we need the names to express arbitrary predicates and to drill down deeper
2. I didn't think on how to represent ALL except of list of all expression results has to be true

but that leads to indexes not being used

we can change this to: `size([(start)-->(next:Next) WHERE next.name = $foo | true]) = 0` where the `= 0` depends on the below

* ANY(some): size > 1
* SINGLE: size = 1
* NONE: size = 0
* ALL: size = 0 with inverted condition

this works a bit for single level but for deeper nesting it breaks the planner

one idea is to skip intermediate levels if there are no additional filters

e.g.  `size([(start)-->(middle)-->(next:Next) WHERE next.name = $foo | true]) = 0`

originally we would generate:

 `size([(start)-->(middle) WHERE size([(middle)-->(next:Next) WHERE next.name = $foo | true]) = 0 | true]) = 0`

in 4.0 we can generate existential subqueries to help with that

`WHERE exists { MATCH ... WHERE ... } ` I think will be the syntax.

jexp:
This still uses the index:

```
profile MATCH (t:User) WITH collect(t) as users
WHERE size([(t)-[:FOLLOWS]->()-[:FOLLOWS]->(u:User) WHERE u.id = "PunkAB" | u]) = 0
RETURN count(*)
```

![plan-nested-index](https://user-images.githubusercontent.com/67427/65592126-45265180-df8e-11e9-8195-ecfed7801b8e.png)

jexp:
This also still uses the index

```
profile MATCH (t:User) with t limit 10
WHERE size([(t)-[:FOLLOWS]->(b:User) WHERE size([(b)-[:FOLLOWS]->(u:User) WHERE u.id = "PunkAB" | true])=0 | true]) = 0
RETURN count(*)

```

![plan-index-2](https://user-images.githubusercontent.com/67427/65592449-ddbcd180-df8e-11e9-9d83-ad9ade0ed987.png)

jexp:
Problem is with the negation for ALL that kills the index :(

```
profile MATCH (t:User) with t limit 10
[...]

PR #78, opened 2020-05-19
Feature/#63 deep filter optimization

Andy2003:
Implements an alternative filtering approach that better supports the internal optimization of cypher queries.

Relates to #63


src/main/java/org/neo4j/opencypherdsl/Aliased.java
```diff
@@ -0,0 +1,50 @@
+/*
```
jexp, line 1: let's not add opencypher-dsl as code but as dependency, release is meant to go out this week.

src/test/resources/optimized-query-for-filter.adoc
```diff
@@ -0,0 +1,897 @@
...
+----
+
+== Tests
+
+=== Filter at same level (some)
+
+.Query configuration
+[source,json,config=true]
+----
+{  "optimizedQuery": true }
```
jexp, line 53: to generic, we should use a specific term here: or an array: `optimizeQuery: ["filter-as-match"]`

src/test/resources/optimized-query-for-filter.adoc
```diff
@@ -0,0 +1,897 @@
...
+
+MATCH (movie)<-[:`ACTED_IN`]-(movie_actor_some:`Person`)
+WHERE movie_actor_some.name = $movie_actor_some_name
+WITH DISTINCT movie
+
+MATCH (movie)<-[:`REVIEWED`]-(movie_reviewedBy_some:`Person`)
+WHERE movie_reviewedBy_some.name = $movie_reviewedBy_some_name
+WITH DISTINCT movie
+
+RETURN DISTINCT movie { .title } AS movie
```
jexp, line 99: don't need the distinct here

src/test/resources/optimized-query-for-filter.adoc
```diff
@@ -0,0 +1,897 @@
...
+.Movie Example result
+|===
+|title
+
+|Cloud Atlas
+|===
+
+=== Filter for relation
+
+CAUTION: *Not yet implemented*
```
jexp, line 111: is this still true?

src/test/resources/optimized-query-for-filter.adoc
```diff
@@ -0,0 +1,897 @@
...
+MATCH (movie)<-[:`ACTED_IN`]-(movie_actor_some:`Person`)
+WHERE
+    movie_actor_some.name = $movie_actor_some_name
+WITH DISTINCT movie
+
+MATCH (movie)<-[:`REVIEWED`]-(movie_reviewedBy_every:`Person`)
+WHERE
+    movie_reviewedBy_every.name = $movie_reviewedBy_every_name
+WITH movie,
+    size((movie)<-[:`REVIEWED`]-(:`Person`)) AS movie_reviewedBy_every_total,
```
jexp, line 228: dann können wir bald auch aggregations implementieren :)

src/test/resources/optimized-query-for-filter.adoc
```diff
@@ -0,0 +1,897 @@
...
+.Cypher
+[source,cypher]
+----
+MATCH (movie:`Movie`)
+WITH movie
+
+MATCH (movie)<-[:`DIRECTED`]-(movie_directedBy_every:`Person`)
+WHERE movie_directedBy_every.name = $movie_directedBy_every_name
+WITH movie,
+    size((movie)<-[:`DIRECTED`]-(:`Person`)) AS movie_directedBy_every_total,
```
jexp, line 288: unfortunately this is not as efficient but probably necessary. as it has to do an expand

jexp: if we would know that there are only Person nodes at the end of DIRECTED we can leave off the label and it becomes a getDegree operation

src/test/resources/optimized-query-for-filter.adoc
```diff
@@ -0,0 +1,897 @@
...
+|===
+|title
+
+|Unforgiven
+|===
+
+
+=== Filter with `OR` and `AND`
+
+CAUTION: *Not yet implemented*
```
jexp, line 317: is this still the case?

src/test/resources/optimized-query-for-filter.adoc
```diff
@@ -0,0 +1,897 @@
...
+  "movie_directedBy_none_name":"Lilly Wachowski"
+}
+----
+
+.Cypher
+[source,cypher]
+----
+MATCH (movie:`Movie`)
+WITH movie
+WHERE size([(movie)<-[:`DIRECTED`]-(movie_directedBy_none:`Person`) WHERE movie_directedBy_none.name = $movie_directedBy_none_name | true]) = 0
```
jexp, line 645: is this the old implementation? the new one would probably use an optional match and a count of 0 ?

src/main/kotlin/org/neo4j/graphql/OptimizedQueryException.kt
```diff
@@ -0,0 +1,3 @@
+package org.neo4j.graphql
+
+class OptimizedQueryException(message: String = "Not implemented, please provide a test case") : Exception(message)
```
jexp, line 3: more detail in the error message?

jexp: *what* is not implemented

src/test/kotlin/org/neo4j/graphql/utils/CypherTestSuite.kt
```diff
@@ -20,16 +20,30 @@ class CypherTestSuite(fileName: String) : AsciiDocTestSuite() {
...
             private val requestParams: Map<String, Any?> = emptyMap(),
             private val ignore: Boolean) {
 
         fun run(contextProvider: ((requestParams: Map<String, Any?>) -> QueryContext?)?) {
             println(title)
             try {
-                val ctx = contextProvider?.let { it(requestParams) } ?: QueryContext()
+                val ctx = contextProvider?.let { it(requestParams) } ?: config ?: QueryContext()
                 val result = suite.translate(request, requestParams, ctx)
+
```
jexp, line 33: isn't that just noise in the test output? if we need it we should have a flag for enabling these outputs

src/main/kotlin/org/neo4j/graphql/QueryContext.kt
```diff
@@ -4,5 +4,10 @@ data class QueryContext @JvmOverloads constructor(
         /**
          * if true the <code>__typename</code> will be always returned for interfaces, no matter if it was queried or not
          */
-        var queryTypeOfInterfaces: Boolean = false
+        var queryTypeOfInterfaces: Boolean = false,
+
+        /**
+         * If true an alternative approach for query translation will be used.
+         */
+        var optimizedQuery: Boolean = false
```
jexp, line 12: change to optimize: [filter-as-match]

src/main/kotlin/org/neo4j/graphql/handler/QueryHandler.kt
```diff
@@ -77,6 +80,18 @@ class QueryHandler private constructor(
         val ordering = orderBy(variable, field.arguments)
         val skipLimit = SkipLimit(variable, field.arguments).format()
 
+        if ((env.getContext() as? QueryContext)?.optimizedQuery == true) {
+
+            val (partialQuery, filterParams) = OptimizedFilterHandler(type).generateFilterQuery(variable, field)
+            val statement = partialQuery
+                ?.returningDistinct(PassThrough("${mapProjection.query} AS $variable$ordering${skipLimit.query}"))
+                ?.build()
+                    ?: throw OptimizedQueryException()
```
jexp, line 89: more detail in the exception

src/main/kotlin/org/neo4j/graphql/handler/QueryHandler.kt
```diff
@@ -77,6 +80,18 @@ class QueryHandler private constructor(
...
+            val query = Renderer.getDefaultRenderer().render(statement)
```
jexp, line 90: should that be something that's shared? perhaps as a config/injection

src/main/kotlin/org/neo4j/graphql/GraphQLExtensions.kt
```diff
@@ -46,13 +46,18 @@ fun GraphQLFieldsContainer.relationshipFor(name: String): RelationshipInfo? {
...
-            ?: field.getDirective(DirectiveConstants.RELATION)?.let { it to false }
-            ?: throw IllegalStateException("Field $field needs an @relation directive")
+    val (relDirective, isRelFromType) = if (isRelationType()){
+        (this as? GraphQLDirectiveContainer)
+            ?.getDirective(DirectiveConstants.RELATION)?.let { it to false }
+                ?: throw IllegalStateException("Field $field needs an @relation directive")
+    } else {
+        (fieldObjectType as? GraphQLDirectiveContainer)
+            ?.getDirective(DirectiveConstants.RELATION)?.let { it to true }
+                ?: field.getDirective(DirectiveConstants.RELATION)?.let { it to false }
```
jexp, line 56: shouldn't in the case that the object type is a relationship-type it fail as well if there is no @relationship directive on the type? it seems to fail only when it's missing on the field, but AFAIR the one on the field is optional then and only needed when it is a same-type relationship, not sure.

src/main/kotlin/org/neo4j/graphql/GraphQLExtensions.kt
```diff
@@ -72,17 +77,17 @@ fun GraphQLFieldsContainer.getValidTypeLabels(schema: GraphQLSchema): List<Strin
...
 
 @Suppress("SimplifiableCallChain")
-fun GraphQLFieldsContainer.label(includeAll: Boolean = false) = when {
+fun GraphQLFieldsContainer.label(includeAll: Boolean = false, quote: Boolean = true) = when {
     this.isRelationType() ->
         (this as? GraphQLDirectiveContainer)
             ?.getDirective(DirectiveConstants.RELATION)
-            ?.getArgument(RELATION_NAME)?.value?.toJavaValue()?.toString()?.quote()
-                ?: this.name.quote()
+            ?.getArgument(RELATION_NAME)?.value?.toJavaValue()?.toString()?.let { if (quote) it.quote() else it  }
```
jexp, line 84: when do we not have the quotes for a label?

btw. odd that in the tck-examples several non-critical labels/rel-types ones were quoted.

jexp: isn't it easier to just have a separate (inner) function that returns the unquoted and then one that wraps it with .quote() this 3 times same condition makes the code less readable and maintainable.

jexp:
We will not copy opencypher-dsl into the library for this PR let's wait for the release and depend on that.

pom.xml
```diff
@@ -22,7 +22,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <java.version>1.8</java.version>
-        <kotlin.version>1.3.41</kotlin.version>
+        <kotlin.version>1.3.72</kotlin.version>
```
jexp, line 25: let's separate changes like this out into separate PRs

pom.xml
```diff
@@ -113,11 +113,14 @@
             <version>${kotlin.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
jexp, line 116: only what we need for the actual PR / change not open-cypher-dsl

src/main/kotlin/org/neo4j/graphql/BuildingEnv.kt
```diff
@@ -92,20 +92,62 @@ class BuildingEnv(val types: MutableMap<String, GraphQLType>) {
                     else -> addFilterType(getInnerFieldsContainer(typeDefinition), createdTypes)
                 }
 
-                Operators.forType(types[filterType] ?: typeDefinition).forEach { op ->
-                    val wrappedType: GraphQLInputType = when {
-                        op.list -> GraphQLList(GraphQLTypeReference(filterType))
-                        else -> GraphQLTypeReference(filterType)
+                if (field.isRelationship()) {
+                    val list = field.type.isList()
```
jexp, line 96: can we extract this huge block into a method?

src/main/kotlin/org/neo4j/graphql/BuildingEnv.kt
```diff
@@ -92,20 +92,62 @@ class BuildingEnv(val types: MutableMap<String, GraphQLType>) {
...
+                    addFilterField(builder, field, RelationOperator.EQ_OR_NOT_EXISTS, filterType,
```
jexp, line 97: can those information bits/methods go into the Operators (or if you want a separate RelationshipOperators) enum: 
- description
- operator
- name

src/main/java/org/neo4j/opencypherdsl/Aliased.java
Andy2003: can you trigger @michael-simons to review + merge https://github.com/neo4j/sdn-rx/pull/244?

src/test/resources/optimized-query-for-filter.adoc
Andy2003: Yes

src/test/resources/optimized-query-for-filter.adoc
Andy2003: This is why the user should always be aware what the `_every` filter means. So we log warnings if he uses it by not defining it explicitly.

src/test/resources/optimized-query-for-filter.adoc
Andy2003: Yes

src/test/resources/optimized-query-for-filter.adoc
Andy2003: :+1:

src/main/kotlin/org/neo4j/graphql/OptimizedQueryException.kt
Andy2003: This is just the default message

src/main/kotlin/org/neo4j/graphql/handler/QueryHandler.kt
Andy2003: It's the only place where it is used, no need to share it by now

src/main/kotlin/org/neo4j/graphql/GraphQLExtensions.kt
Andy2003: This is the old logic, I just added the branch for handling relation types. I'll adjust the message of the Exception for the new branch.

pom.xml
Andy2003: I will remove as soon we get an release of `open-cypher-dsl`

src/main/kotlin/org/neo4j/graphql/BuildingEnv.kt
Andy2003: since there the descriptions are not static (b/c of `type` and `name`) and there are different descriptions for `n..1` and `n..m`, I would like to keep it in a method.

src/main/kotlin/org/neo4j/graphql/Predicates.kt
```diff
@@ -4,22 +4,33 @@ import graphql.Scalars
...
 
     companion object {
         fun resolvePredicate(name: String, value: Any?, type: GraphQLFieldsContainer): Predicate {
-            val (fieldName, op) = Operators.resolve(name, value)
-            return if (type.hasRelationship(fieldName)) {
-                when (value) {
-                    is Map<*, *> -> RelationPredicate(fieldName, op, value, type)
-                    null -> IsNullPredicate(fieldName, op, type)
-                    else -> throw IllegalArgumentException("Input for $fieldName must be an filter-InputType")
+            for (definedField in type.fieldDefinitions) {
```
jexp, line 16: looks more expensive, why is this necessary to iterate over all fields?

src/main/kotlin/org/neo4j/graphql/Predicates.kt
```diff
@@ -59,21 +70,21 @@ data class CompoundPredicate(val parts: List<Predicate>, val op: String = "AND")
                 }
 }
 
-data class IsNullPredicate(val fieldName: String, val op: Operators, val type: GraphQLFieldsContainer) : Predicate {
+data class IsNullPredicate(val fieldName: String, val op: RelationOperator, val type: GraphQLFieldsContainer) : Predicate {
```
jexp, line 73: what about the other, non-relationship is null predicates?

src/main/kotlin/org/neo4j/graphql/Predicates.kt
```diff
@@ -86,18 +97,34 @@ data class ExpressionPredicate(val name: String, val op: Operators, val value: A
...
+        if (type.getFieldDefinition(fieldName).isList()) {
+            if (op == RelationOperator.EQ_OR_NOT_EXISTS) {
+                LOGGER.info("$fieldName on type ${type.name} was used for filtering, consider using ${fieldName}${RelationOperator.EVERY.suffix} instead")
+            }
+        } else {
+            @Suppress("NON_EXHAUSTIVE_WHEN")
+            when (op) {
+                RelationOperator.SINGLE -> LOGGER.warn("Using $fieldName${RelationOperator.SINGLE.suffix} on type ${type.name} is deprecated, use ${fieldName} directly")
+                RelationOperator.SOME -> LOGGER.warn("Using $fieldName${RelationOperator.SOME.suffix} on type ${type.name} is deprecated, use ${fieldName} directly")
+                RelationOperator.NONE -> LOGGER.warn("Using $fieldName${RelationOperator.NONE.suffix} on type ${type.name} is deprecated, use ${fieldName}${RelationOperator.NOT.suffix} instead")
```
jexp, line 125: why don't we handle the other cases? i.e. with an error?

src/main/kotlin/org/neo4j/graphql/Predicates.kt
```diff
@@ -4,22 +4,33 @@ import graphql.Scalars
...
         fun resolvePredicate(name: String, value: Any?, type: GraphQLFieldsContainer): Predicate {
-            val (fieldName, op) = Operators.resolve(name, value)
-            return if (type.hasRelationship(fieldName)) {
-                when (value) {
-                    is Map<*, *> -> RelationPredicate(fieldName, op, value, type)
-                    null -> IsNullPredicate(fieldName, op, type)
-                    else -> throw IllegalArgumentException("Input for $fieldName must be an filter-InputType")
+            for (definedField in type.fieldDefinitions) {
+                if (definedField.isRelationship()) {
+                    val op = RelationOperator.values().find { name == definedField.name + it.suffix }
```
jexp, line 18: enumValues<RelationOperator>

src/main/kotlin/org/neo4j/graphql/Predicates.kt
```diff
@@ -107,67 +134,45 @@ data class RelationPredicate(val fieldName: String, val op: Operators, val value
...
+            return op
+        }
 
-        fun forType(type: GraphQLType): List<Operators> =
+        fun forType(type: GraphQLType): List<FieldOperator> =
                 when {
                     type == Scalars.GraphQLBoolean -> listOf(EQ, NEQ)
                     type.isNeo4jType() -> listOf(EQ, NEQ, IN, NIN)
-                    type is GraphQLFieldsContainer || type is GraphQLInputObjectType -> listOf(IS_NULL, IS_NOT_NULL, SOME, NONE, SINGLE)
+                    type is GraphQLFieldsContainer || type is GraphQLInputObjectType -> throw IllegalArgumentException("This operators are not for relations, use the RelationOperator instead")
```
jexp, line 175: what should user do with this error message?

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
+package org.neo4j.graphql.handler.filter
```
jexp, line 1: pretty large file with pretty large functions, can we split this up a bit more to make it more understandable?

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+import org.neo4j.opencypherdsl.*
+import org.neo4j.opencypherdsl.Cypher
+import org.neo4j.opencypherdsl.Node
+import org.slf4j.LoggerFactory
+
+class OptimizedFilterHandler(val type: GraphQLFieldsContainer) {
+
+    fun generateFilterQuery(variable: String, field: Field): Pair<StatementBuilder.OrderableOngoingReadingAndWithWithoutWhere?, MutableMap<String, Any?>> {
+        if (type.isRelationType()) {
+            throw OptimizedQueryException("Optimization for relation not implemented, please provide a test case")
```
jexp: for relationship entity type

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+        fun parseFilter(additionalFilter: ((StatementBuilder.ExposesWith) -> StatementBuilder.OrderableOngoingReadingAndWithWithoutWhere)? = null): StatementBuilder.OrderableOngoingReadingAndWithWithoutWhere {
+            if (value !is ObjectValue) {
+                throw OptimizedQueryException()
+            }
+            // WHERE MATCH all predicates for current
+            // WITH x
+            var query = addWhere(additionalFilter)
+
+            // Handle all quantifier (ALL / NONE / SINGLE / ANY)
+            query = handleRelation(query)
```
jexp, line 73: handleQuantifier?

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                additionalFilter: ((StatementBuilder.ExposesWith) -> StatementBuilder.OrderableOngoingReadingAndWithWithoutWhere)? = null
+        ): StatementBuilder.OrderableOngoingReadingAndWithWithoutWhere {
+            val reading = readingWithoutWhereFactory()
+            if (parsedQuery.fieldPredicates.isEmpty()) {
+                if (reading is StatementBuilder.OrderableOngoingReadingAndWithWithoutWhere) {
+                    return reading
+                }
+            }
+            if (reading !is StatementBuilder.OngoingReadingWithoutWhere) {
+                throw IllegalStateException("cannot handle " + reading::class.java)
```
jexp, line 89: ??

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+            // Handle OR / AND
+            return handleCombinations(query)
+        }
+
+        private fun addWhere(
+                additionalFilter: ((StatementBuilder.ExposesWith) -> StatementBuilder.OrderableOngoingReadingAndWithWithoutWhere)? = null
+        ): StatementBuilder.OrderableOngoingReadingAndWithWithoutWhere {
+            val reading = readingWithoutWhereFactory()
+            if (parsedQuery.fieldPredicates.isEmpty()) {
+                if (reading is StatementBuilder.OrderableOngoingReadingAndWithWithoutWhere) {
```
jexp, line 84: can we give these predicates a bit more meaning?, e.g with a named variable

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+            query = handleRelation(query)
+
+            // Handle OR / AND
+            return handleCombinations(query)
+        }
+
+        private fun addWhere(
+                additionalFilter: ((StatementBuilder.ExposesWith) -> StatementBuilder.OrderableOngoingReadingAndWithWithoutWhere)? = null
+        ): StatementBuilder.OrderableOngoingReadingAndWithWithoutWhere {
+            val reading = readingWithoutWhereFactory()
```
jexp, line 82: reading is an odd name,  can we find something better that resonates with what it is?

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                val withs = if (parsedQuery.relationPredicates.isNotEmpty() && parentPassThroughWiths.firstOrNull { it == current.requiredSymbolicName } == null) {
+                    parentPassThroughWiths + current.requiredSymbolicName
+                } else {
+                    parentPassThroughWiths
+                }
+                withClauseWithOptionalDistinct(readingWithWhere ?: reading, withs, useDistinct)
+            }
+        }
+
+        private fun handleRelation(passedQuery: StatementBuilder.OrderableOngoingReadingAndWithWithoutWhere): StatementBuilder.OrderableOngoingReadingAndWithWithoutWhere {
```
jexp, line 108: this function is definitely too long and needs to be broken up into a series of steps

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                    val where = when (op) {
+                        RelationOperator.NOT -> query.where(relation)
+                        RelationOperator.EQ_OR_NOT_EXISTS -> query.where(Conditions.not(relation))
+                        else -> throw IllegalStateException("$op should not be set for Null value")
+                    }
+                    query = withClauseWithOptionalDistinct(where, levelPassThroughWiths)
+                    continue
+                }
+                if (objectField.value !is ObjectValue) {
+                    throw OptimizedQueryException()
```
jexp: concrete error message

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+
+                val (op, objectField, relField) = relFilter
+                val rel = type.relationshipFor(relField.name)!!
+                val label = (relField.type.inner() as? GraphQLObjectType)?.label(quote = false)!!
+                val relNode = Cypher.node(label)
+
+                if (objectField.value is NullValue) {
+                    // EXISTS + NOT EXISTS
+                    if (current !is Node) {
+                        throw OptimizedQueryException()
```
jexp: concrete error message

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                val relVariableName = variablePrefix + "_" + objectField.name
+                val relVariable = relNode.named(relVariableName)
+
+                if (index + 1 < parsedQuery.relationPredicates.size) {
+                    // if there are additional relationPredicates, we need to pass through the current
+                    levelPassThroughWiths.add(current.requiredSymbolicName)
+                }
+
+                val nestedRelationship = { target: Node ->
+                    createRelation(rel, current as? Node ?: throw OptimizedQueryException(), target)
```
jexp: concrete error message

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                        relVariable,
+                        relVariableName,
+                        { if (hasPredicates) query.match(nestedRelationship(relVariable)) else query },
+                        rel.type,
+                        objectField.value,
+                        filterParams,
+                        levelPassThroughWiths)
+
+                // used as additional filter to handle SINGLE and EVERY use cases
+                val totalAndCountFilter = { whereClause: (withWithoutWhere: StatementBuilder.OrderableOngoingReadingAndWithWithoutWhere, totalVar: SymbolicName, countVar: SymbolicName) -> StatementBuilder.OrderableOngoingReadingAndWithWithWhere ->
```
jexp: separate method

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                        val withWithoutWhere = withClauseWithOptionalDistinct(
+                                exposesWith,
+                                levelPassThroughWiths + additionalWiths + listOf(total, count)
+                        )
+                        val where = whereClause(withWithoutWhere, Cypher.name(totalVar), Cypher.name(countVar))
+                        withClauseWithOptionalDistinct(where, levelPassThroughWiths + additionalWiths)
+                    }
+                }
+
+                when (op) {
```
jexp, line 208: sparate function

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                    RelationOperator.EVERY -> query = nestingLevelHandler.parseFilter(totalAndCountFilter({ withWithoutWhere, total, count ->
+                        withWithoutWhere.where(total.isEqualTo(count))
+                    }))
+                    RelationOperator.SINGLE -> query = nestingLevelHandler.parseFilter(totalAndCountFilter({ withWithoutWhere, total, count ->
+                        withWithoutWhere
+                            .where(total.isEqualTo(count))
+                            .and(total.isEqualTo(Cypher.literalOf(1)))
+                    }))
+                    RelationOperator.NONE -> {
+                        val patternWithoutWhere = Cypher.listBasedOn(createRelation(rel, current as? Node
```
jexp: can we pull this into the if?

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                                ?: throw OptimizedQueryException(), relVariable))
```
jexp: error message

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                                val patternWithoutReturn = addConditions<PatternComprehension.OngoingDefinitionWithoutReturn>(
+                                        relVariable, relVariableName, nestedParsedQuery.fieldPredicates, filterParams) { where, condition ->
+                                    where?.and(condition) ?: patternWithoutWhere.where(condition)
+                                } ?: patternWithoutWhere
+                                val where = query
+                                    .where(Functions.size(patternWithoutReturn.returning(Cypher.literalTrue())).isEqualTo(Cypher.literalOf(0)))
+                                query = withClauseWithOptionalDistinct(where, levelPassThroughWiths)
+                            }
+                        } else {
+                            throw OptimizedQueryException()
```
jexp: error message

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                        }
+                    }
+                    else -> throw IllegalStateException("$op should not be set for relationPredicates")
```
jexp, line 225: add type/field name?

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                    RelationOperator.SINGLE -> query = nestingLevelHandler.parseFilter(totalAndCountFilter({ withWithoutWhere, total, count ->
+                        withWithoutWhere
+                            .where(total.isEqualTo(count))
+                            .and(total.isEqualTo(Cypher.literalOf(1)))
+                    }))
+                    RelationOperator.NONE -> {
+                        val patternWithoutWhere = Cypher.listBasedOn(createRelation(rel, current as? Node
+                                ?: throw OptimizedQueryException(), relVariable))
+                        if (nestedParsedQuery.and.isNullOrEmpty() && nestedParsedQuery.or.isNullOrEmpty() && nestedParsedQuery.relationPredicates.isNullOrEmpty()) {
+                            if (nestedParsedQuery.fieldPredicates.isNotEmpty()) {
```
jexp: can we simplify this bit, it looks super complicated

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                }
+            }
+            return query
+        }
+
+        private fun handleCombinations(passedQuery: StatementBuilder.OrderableOngoingReadingAndWithWithoutWhere): StatementBuilder.OrderableOngoingReadingAndWithWithoutWhere {
+            if ( parsedQuery.or.isNullOrEmpty() && parsedQuery.and.isNullOrEmpty()){
+                return passedQuery
+            }
+            throw OptimizedQueryException()
```
jexp: concrete error message

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                        }
+                    }
+                    else -> throw IllegalStateException("$op should not be set for relationPredicates")
+                }
+            }
+            return query
+        }
+
+        private fun handleCombinations(passedQuery: StatementBuilder.OrderableOngoingReadingAndWithWithoutWhere): StatementBuilder.OrderableOngoingReadingAndWithWithoutWhere {
+            if ( parsedQuery.or.isNullOrEmpty() && parsedQuery.and.isNullOrEmpty()){
```
jexp: I think we had that predicate already before

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                return passedQuery
+            }
+            throw OptimizedQueryException()
+        }
+
+        private fun withClauseWithOptionalDistinct(
+                exposesWith: StatementBuilder.ExposesWith,
+                withs: Collection<Expression>,
+                useDistinct: Boolean = true): StatementBuilder.OrderableOngoingReadingAndWithWithoutWhere =
```
jexp, line 241: does this work without the type def?

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+            val relationPredicates = mutableListOf<Predicate<RelationOperator>>()
+            val or: List<Value<*>>?
+            val and: List<Value<*>>?
+
+            init {
+                val queriedFields = objectValue.objectFields
+                    .mapIndexed { index, field -> field.name to (index to field) }
+                    .toMap(mutableMapOf())
+                or = queriedFields.remove("OR")?.second?.value?.let {
+                    return@let (it as ArrayValue).values
```
jexp, line 295: not sure we need the return@let here. 
This should work too: `value?.let { if (false) "a" else null} ?: throw RuntimeException()`
there should be a shorter way with a conditional let but I don't have an IDE running currently.

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                            ?: throw IllegalArgumentException("OR is expected to be a list")
```
jexp, line 296: provide more detail in the error message like type and field

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                }
+                and = queriedFields.remove("AND")?.second?.value?.let {
```
jexp, line 298: same as above

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                    return@let (it as ArrayValue).values
+                            ?: throw IllegalArgumentException("AND is expected to be a list")
+                }
+                for (definedField in type.fieldDefinitions) {
+                    if (definedField.isRelationship()) {
+                        collectRelationPredicates(type, definedField, queriedFields)
```
jexp, line 304: is this recursive? 
could we avoid passing in a mutable map?
and instead use the return values (e.g. keys) to transform the map?

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                    } else {
+                        collectFieldPredicates(definedField, queriedFields)
```
jexp, line 306: same here

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                    }
+                }
+                if (queriedFields.isNotEmpty()) {
+                    throw OptimizedQueryException("queried unknown fields: " + queriedFields.keys)
```
jexp: more detail type+fields

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                    .toMap(mutableMapOf())
+                or = queriedFields.remove("OR")?.second?.value?.let {
+                    return@let (it as ArrayValue).values
+                            ?: throw IllegalArgumentException("OR is expected to be a list")
+                }
+                and = queriedFields.remove("AND")?.second?.value?.let {
+                    return@let (it as ArrayValue).values
+                            ?: throw IllegalArgumentException("AND is expected to be a list")
+                }
+                for (definedField in type.fieldDefinitions) {
```
jexp, line 302: extract into method?

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                    }
+                }
+                if (queriedFields.isNotEmpty()) {
+                    throw OptimizedQueryException("queried unknown fields: " + queriedFields.keys)
+                }
+                fieldPredicates.sortedBy(Predicate<FieldOperator>::index)
+                relationPredicates.sortedBy(Predicate<RelationOperator>::index)
+            }
+
+            private fun collectRelationPredicates(type: GraphQLFieldsContainer, definedField: GraphQLFieldDefinition, queriedFields: MutableMap<String, Pair<Int, ObjectField>>) {
```
jexp, line 316: this is a pretty huge method, can we split it up,  and perhaps move stuff into the Operator?

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                        val op = when (definedField.type.isList()) {
+                            true -> when (queryOp) {
+                                RelationOperator.NOT -> when (objectField.value) {
+                                    is NullValue -> RelationOperator.NOT
+                                    else -> RelationOperator.NONE
+                                }
+                                RelationOperator.EQ_OR_NOT_EXISTS -> when (objectField.value) {
+                                    is NullValue -> RelationOperator.EQ_OR_NOT_EXISTS
+                                    else -> {
+                                        LOGGER.info("$queryFieldName on type ${type.name} was used for filtering, consider using ${definedField.name}${RelationOperator.EVERY.suffix} instead")
```
jexp, line 330: this would be really spammy in the logs, as it shows up every time, consider `debug` instead

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                                        RelationOperator.EVERY
+                                    }
+                                }
+                                else -> queryOp
+                            }
+                            false -> when (queryOp) {
+                                RelationOperator.SINGLE -> {
+                                    LOGGER.warn("Using $queryFieldName on type ${type.name} is deprecated, use ${definedField.name} directly")
```
jexp, line 338: this would be really spammy in the logs, as it shows up every time, consider `debug` instead

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                fieldPredicates.sortedBy(Predicate<FieldOperator>::index)
+                relationPredicates.sortedBy(Predicate<RelationOperator>::index)
+            }
+
+            private fun collectRelationPredicates(type: GraphQLFieldsContainer, definedField: GraphQLFieldDefinition, queriedFields: MutableMap<String, Pair<Int, ObjectField>>) {
+                RelationOperator.values()
+                    .map { it to definedField.name + it.suffix }
+                    .mapNotNull { (queryOp, queryFieldName) ->
+                        val (index, objectField) = queriedFields.remove(queryFieldName) ?: return@mapNotNull null
+                        val op = when (definedField.type.isList()) {
```
jexp, line 321: extract branches into methods or better move into RelationOperator.transformPredicate(isList)

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                        Predicate(op, objectField, definedField, index)
+                    }
+                    .forEach { relationPredicates.add(it) }
+            }
+
+            private fun collectFieldPredicates(definedField: GraphQLFieldDefinition, queriedFields: MutableMap<String, Pair<Int, ObjectField>>) {
+                FieldOperator.values()
+                    .map { it to definedField.name + it.suffix }
+                    .mapNotNull { (predicate, queryFieldName) ->
+                        val (index, objectField) = queriedFields.remove(queryFieldName) ?: return@mapNotNull null
```
jexp, line 369: again, is there a better way to do this without return@ and mutating the map here?
and without mutating fieldPredicates

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -0,0 +1,369 @@
...
+                                RelationOperator.EQ_OR_NOT_EXISTS -> when (objectField.value) {
+                                    is NullValue -> RelationOperator.EQ_OR_NOT_EXISTS
+                                    else -> RelationOperator.SOME
+                                }
+                                else -> queryOp
+                            }
+                        }
+                        Predicate(op, objectField, definedField, index)
+                    }
+                    .forEach { relationPredicates.add(it) }
```
jexp, line 362: again, is there a better way to do this without return@ and mutating the map here?
and without mutating relPredicates

src/main/kotlin/org/neo4j/graphql/BuildingEnv.kt
```diff
@@ -92,20 +92,66 @@ class BuildingEnv(val types: MutableMap<String, GraphQLType>) {
...
                 }
+
             }
         types[filterName] = builder.build()
         return filterName
     }
 
+    private fun createRelationFilterFields(type: GraphQLFieldsContainer, field: GraphQLFieldDefinition, filterType: String, builder: GraphQLInputObjectType.Builder) {
+        val list = field.type.isList()
+        addFilterField(builder, field, RelationOperator.EQ_OR_NOT_EXISTS, filterType,
```
jexp, line 109: I think more of this functionality, incl the descriptions and the add field to build could be moved into RelOperator

jexp: So it's all in the same place and not spread throughout the code

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
Andy2003: we cannot pull the `throws` out of the let since this would change the logic

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
Andy2003: So the user will only notice after he enable debug log for this library

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
Andy2003: moved into `RelationOperator`

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
Andy2003: since we need to remove the matching entries from the `queriedFields` map, we need to keep it all together at one place.

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
Andy2003: Nope

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
Andy2003: Adjusted, but should never be thrown

src/main/kotlin/org/neo4j/graphql/Predicates.kt
Andy2003: This is nothing the user will see, it is an error in the implementation

src/main/kotlin/org/neo4j/graphql/Predicates.kt
Andy2003: ??

src/main/kotlin/org/neo4j/graphql/Predicates.kt
Andy2003: b/c it is not an error

src/main/kotlin/org/neo4j/graphql/Predicates.kt
Andy2003: We use different resolve logic for relations than for fields, so we need to know which logic to take

src/main/kotlin/org/neo4j/graphql/BuildingEnv.kt
jexp: ??

src/main/kotlin/org/neo4j/graphql/Predicates.kt
jexp: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/enum-values.html

Andy2003: What is the benefit over using `values()` in the current use case?

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -9,13 +9,18 @@ import org.neo4j.graphql.handler.projection.ProjectionBase
 import org.neo4j.opencypherdsl.*
 import org.neo4j.opencypherdsl.Cypher
 import org.neo4j.opencypherdsl.Node
-import org.slf4j.LoggerFactory
+
+
+typealias WhereClauseFactory = (
```
jexp, line 14: please add type aliases also for the other long names from the DSL, makes it much easier to read
or at least an static import

src/main/kotlin/org/neo4j/graphql/Predicates.kt
jexp: idiomatic language use, but it's not crucial

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
jexp: actually what we should do is to provide the info as part of the translation response so that the middleware server can send it to the client. the server logs won't be read by the folks sending the queries anyway :)

Andy2003: :+1:

Andy2003: I created issue https://github.com/neo4j-graphql/neo4j-graphql-java/issues/80 for this

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
jexp: I would want to revisit that, I think we can have a transformation instead, but that's something we can look at later

PR #94, opened 2020-07-16
Add support for spatial point values and distance comparisons

### LINKED ISSUE #11: support spatial
support spatial point value

what about spatial operations like distance?
- first only via `@cypher` directive

Andy2003:
This PR add support for [spatials](https://neo4j.com/docs/cypher-manual/3.5/syntax/spatial/#cypher-spatial-crs-geographic) and [distance comparisons](https://neo4j.com/docs/cypher-manual/3.5/functions/spatial/#functions-distance).

Now you can rund queries like:

```graphql
query {
  person(location:{longitude: 1, latitude: 2 }){
    name
    location {
      crs
      longitude
      latitude
      height
    }
  }
}
```

or 

```graphql
{
  person(filter:{ location_distance_gte: { distance: 3, point: {longitude: 1, latitude:2, height: 3}}}){
    name
  }
}
```
resolves #11

### DIFF
--- a/src/main/kotlin/org/neo4j/graphql/Predicates.kt
+++ b/src/main/kotlin/org/neo4j/graphql/Predicates.kt
@@ -6,9 +6,10 @@ import graphql.language.Value
 import graphql.schema.*
 import org.neo4j.graphql.Predicate.Companion.resolvePredicate
 import org.neo4j.graphql.handler.projection.ProjectionBase
-import org.neo4j.opencypherdsl.Condition
-import org.neo4j.opencypherdsl.Expression
+import org.neo4j.opencypherdsl.*
 import org.slf4j.LoggerFactory
+
+typealias CypherDSL = org.neo4j.opencypherdsl.Cypher
 
 interface Predicate {
     fun toExpression(variable: String): Cypher
@@ -21,14 +22,26 @@ interface Predicate {
                     if (op != null) {
                         return when (value) {
                             is Map<*, *> -> RelationPredicate(definedField.name, op, value, type)
-                            null -> IsNullPredicate(definedField.name, op, type)
+                            null -> IsNullRelationPredicate(definedField.name, op, type)
                             else -> throw IllegalArgumentException("Input for ${definedField.name} must be an filter-InputType")
                         }
                     }
                 } else {
-                    val op = FieldOperator.resolve(name, definedField.name, value)
+                    val op = FieldOperator.resolve(name, definedField, value)
                     if (op != null) {
-                        return ExpressionPredicate(definedField.name, op, value, definedField)
+                        if (definedField.type.isNeo4jType()) {
+                            if (op.distance) {
+                                return DistancePredicate(definedField.name, op, value, definedField)
+                            }
+                            return when (value) {
+                                is Map<*, *> -> CompoundPredicate(value.entries.map { (key, value) ->
+                                    ExpressionPredicate(definedField.name, key as String, op, value, definedField)
+                                })
+                                null -> IsNullFieldPredicate(definedField.name, op, type)
+                                else -> throw IllegalArgumentException("Input for ${definedField.name} must be an filter-InputType")
+                            }
+                        }
+                        return ExpressionPredicate(definedField.name, null, op, value, definedField)
                     }
                 }
             }
@@ -72,7 +85,7 @@ data class CompoundPredicate(val parts: List<Predicate>, val op: String = "AND")
                 }
 }
 
-data class IsNullPredicate(val fieldName: String, val op: RelationOperator, val type: GraphQLFieldsContainer) : Predicate {
+data class IsNullRelationPredicate(val fieldName: String, val op: RelationOperator, val type: GraphQLFieldsContainer) : Predicate {
     override fun toExpression(variable: String): Cypher {
         val rel = type.relationshipFor(fieldName) ?: throw IllegalArgumentException("Not a relation")
         val (left, right) = rel.arrows
@@ -81,10 +94,23 @@ data class IsNullPredicate(val fieldName: String, val op: RelationOperator, val 
     }
 }
 
-data class ExpressionPredicate(val name: String, val op: FieldOperator, val value: Any?, val fieldDefinition: GraphQLFieldDefinition) : Predicate {
+data class IsNullFieldPredicate(val fieldName: String, val op: FieldOperator, val type: GraphQLFieldsContainer) : Predicate {
+    override fun toExpression(variable: String): Cypher {
+        val check = if (op == FieldOperator.IS_NOT_NULL) "IS NOT NULL" else "IS NULL"
+        return Cypher("$variable.$fieldName $check")
+    }
+}
+
+data class ExpressionPredicate(
+        val name: String,
+        val nestedField: String? = null,
+        val op: FieldOperator,
+        val value: Any?,
+        val fieldDefinition: GraphQLFieldDefinition
+) : Predicate {
     val not = if (op.not) "NOT " else ""
     override fun toExpression(variable: String): Cypher {
-        val paramName: String = ProjectionBase.FILTER + paramName(variable, name, value).capitalize() + "_" + op.name
+        var paramName: String = ProjectionBase.FILTER + paramName(variable, name, value).capitalize() + "_" + op.name
         val query = if (fieldDefinition.isNativeId()) {
             if (op.list) {
                 "${not}ID($variable) ${op.op} [id IN \$$paramName | toInteger(id)]"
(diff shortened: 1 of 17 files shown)


michael-simons:
FYI Cypher-DSL 2020.0.1 with your PR should be on central soonish.

```
<dependency>
        <groupId>org.neo4j</groupId>
        <artifactId>neo4j-cypher-dsl</artifactId>
        <version>2020.0.1</version>
</dependency>
```

src/main/kotlin/org/neo4j/graphql/Predicates.kt
```diff
@@ -21,14 +26,26 @@ interface Predicate {
...
+                            null -> IsNullRelationPredicate(definedField.name, op, type)
                             else -> throw IllegalArgumentException("Input for ${definedField.name} must be an filter-InputType")
                         }
                     }
                 } else {
-                    val op = FieldOperator.resolve(name, definedField.name, value)
+                    val op = FieldOperator.resolve(name, definedField, value)
                     if (op != null) {
-                        return ExpressionPredicate(definedField.name, op, value, definedField)
+                        if (definedField.type.isNeo4jType()) {
```
jexp, line 36: can we extract the block into a method?

src/main/kotlin/org/neo4j/graphql/Predicates.kt
```diff
@@ -81,23 +98,48 @@ data class IsNullPredicate(val fieldName: String, val op: RelationOperator, val
...
             if (op.list) {
                 "${not}ID($variable) ${op.op} [id IN \$$paramName | toInteger(id)]"
             } else {
                 "${not}ID($variable) ${op.op} toInteger(\$$paramName)"
             }
         } else {
-            "$not$variable.${name.quote()} ${op.op} \$$paramName"
+            val appendField = if (nestedField != null) {
+                paramName += "_$nestedField"
+                ".$nestedField"
```
jexp, line 127: paramNames can't contain dots ?
is there a better way than modfiying the paramName?

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -260,15 +262,32 @@ class OptimizedFilterHandler(val type: GraphQLFieldsContainer) {
...
                 conditionAdder: (where: WithWhere?, condition: Condition) -> WithWhere
         ): WithWhere? {
             var where: WithWhere? = null
             for (conditionField in conditions) {
-                val prop = propertyContainer.property(conditionField.fieldDefinition.name)
-                val parameter = Cypher.parameter(variablePrefix + "_" + conditionField.queryField.name)
-                val condition = conditionField.op.conditionCreator(prop, parameter)
-                filterParams[parameter.name] = conditionField.queryField.value.toJavaValue()
-                where = conditionAdder(where, condition)
+                if (conditionField.fieldDefinition.type.isNeo4jType()
```
jexp, line 270: we have a lot of places where larger chunks of code related to spatial / distance are now spread through the code-base like here, would be there any way of extracting them/putting them together so that the implementation doesn't leak into all these places?

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -260,15 +262,32 @@ class OptimizedFilterHandler(val type: GraphQLFieldsContainer) {
...
+                    && !conditionField.op.distance
+                ) {
+                    val nestedConditions = (conditionField.queryField.value as ObjectValue).objectFields
+                        .mapIndexed { index, objectField -> Predicate(conditionField.op, objectField, conditionField.fieldDefinition, index) }
+                    addConditions(propertyContainer, variablePrefix + "_" + conditionField.queryField.name,
+                            nestedConditions, filterParams, true, conditionAdder)
+                } else {
+                    val parameter = Cypher.parameter(variablePrefix + "_" + conditionField.queryField.name)
+                    var name = conditionField.fieldDefinition.name
+                    if (appendObjectFieldName) {
```
jexp, line 281: also more boolean conditions like this one that are passed through which make the code much harder to understand and maintain.

jexp: esp. as they only communicate the what but not the why and what does it mean.

jexp: it would also be good if we found a way to co-locate these code-chunks for the different operations (mostly expressions) with the enums / expression types that were originally meant for keeping that functionality isolated.

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -340,8 +359,16 @@ class OptimizedFilterHandler(val type: GraphQLFieldsContainer) {
                         FieldOperator.values()
                             .map { it to definedField.name + it.suffix }
                             .mapNotNull { (predicate, queryFieldName) ->
-                                queriedFields.remove(queryFieldName)?.let { (index, objectField) ->
-                                    Predicate(predicate, objectField, definedField, index)
+                                queriedFields[queryFieldName]?.let { (index, objectField) ->
+                                    if (predicate.requireParam xor (objectField.value !is NullValue)) {
```
jexp, line 363: first time I've seen an XOR in real code. esp. xor with negations is kinda ouch

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
```diff
@@ -340,8 +359,16 @@ class OptimizedFilterHandler(val type: GraphQLFieldsContainer) {
...
+                                        // if we got a value but the predicate requires none
```
jexp, line 364: can we turn the comment into code instead?

src/test/kotlin/org/neo4j/graphql/utils/AsciiDocTestSuite.kt
```diff
@@ -72,16 +72,14 @@ open class AsciiDocTestSuite {
         }
 
         private fun fixNumber(v: Any?): Any? = when (v) {
-            is Float -> v.toDouble(); is Int -> v.toLong(); else -> v
+            is Float -> v.toDouble()
+            is Int -> v.toLong()
+            is List<*> -> v.map { fixNumber(it) }
```
jexp, line 77: probably for Iterable/Sequence?

src/test/resources/augmentation-tests.adoc
```diff
@@ -214,6 +229,42 @@ type _Neo4jLocalTime {
   second: Int
 }
 
+type _Neo4jPoint {
```
jexp, line 232: I wonder if we can minimize that.

Either by splitting it into 2 point types.
And removing / limiting a bit of the capabilities (e.g. coordinate systems) ?

/cc @johnymontana

johnymontana: This looks to match the point type used in neo4j-graphql.js, which is good. I don't think we want multiple GraphQL point types, and instead we should match the Neo4j database point type.

I think the issue @jexp is getting at here is that there are many optional fields on the point type. This stems from the fact that there aren't separate database types for 2D, or 3D - instead they are the same type. So to match that behavior the GraphQL type needs to have these optional fields. One thing that could be done in the future to simplify this is to use a GraphQL union input type (coming to the GraphQL spec soon) for the mutations and filters, which would have a LatLon, xyz, etc types wrapped up in a point union input type which I think would make it more clear what the various options for specifying a point type are.

In most cases I think users are interested in latitude, longitude with the geographic CRS, which is the default.

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
Andy2003: I did this before the `xor` but the condition was even uglier than the xor. Do you really don't like the `xor` so much.
In my opinion this line is even worse:
```kotlin
if ((predicate.requireParam && objectField.value is NullValue) || (!predicate.requireParam && objectField.value !is NullValue))
```

src/main/kotlin/org/neo4j/graphql/handler/filter/OptimizedFilterHandler.kt
Andy2003: :-)

src/main/kotlin/org/neo4j/graphql/Predicates.kt
Andy2003: they can, but it breaks the "style" so I've chosen to use underscore

Andy2003: they can't, since neo4j will try to resolve everything behind the `.` as property of a scalar type

PR #96, opened 2020-07-22
Fixes the issue that an update removes the custom ID field on the node

### LINKED ISSUE #95: Update mutation removes custom ID field from translated Cypher
Since upgrading to version `1.0.0` of the library, I've noticed that the update mutations that we translate to Cypher are not including our custom 'ID' type field in the SET statement. For various reasons, we need to use a separately declared 'ID' field that isn't `_id`.

The effect of this is that the `ID` field is dropped from the resulting object in the Neo4j database.

Here's a simplified example based on our current usecase:

Schema:
```
type Person {
    userId : ID!
    username: String
}
```

Input GraphQL:
```
mutation {
    updatePerson(username: $username, userId: $userId) { username userId } 
}
```

Output Cypher:
```
MATCH (updateParson:Person { userId: $userId }) SET updatePerson =  { username: $username } WITH updatePerson RETURN updatePerson { .username, .userId } AS updatePerson
```

Relevant Java translation code:
```
final GraphQLSchema graphQLSchema = SchemaBuilder.buildSchema(schema);
translator = new Translator(graphQLSchema);
final List<Cypher> cypher = translator.translate(query, variables);
```

Pre-update Person:
```
{
    userId: "<SOME USER ID>",
    username: "John.Smith"
}
```

Post-update Person:
```
{
    username: "Jeremy.Smith"
}
```

This used to work correctly in `1.0.0-M03` (see relevant code [here](https://github.com/neo4j-graphql/neo4j-graphql-java/blob/1.0.0-M03/src/main/kotlin/org/neo4j/graphql/Translator.kt#L117-L119)), where the userId would correctly be passed through to the resulting Cypher.

In version `1.0.0`, the Cypher is built [here](https://github.com/neo4j-graphql/neo4j-graphql-java/blob/1.0.0/src/main/kotlin/org/neo4j/graphql/Translator.kt#L84-L92). It doesn't explicitly exclude the `userId` field, but I have a feeling that it's related to the fact that it's an `ID` type field.

This issue could possibly be related to [issue 83](https://github.com/neo4j-graphql/neo4j-graphql-java/issues/83).

Andy2003:
We forgot to pass the id, when updating a node. This causes the id to be unset.
This PR resolves #95

### DIFF
--- a/src/main/kotlin/org/neo4j/graphql/handler/MergeOrUpdateHandler.kt
+++ b/src/main/kotlin/org/neo4j/graphql/handler/MergeOrUpdateHandler.kt
@@ -31,7 +31,7 @@ class MergeOrUpdateHandler private constructor(
         }
 
         override fun createDataFetcher(rootType: GraphQLObjectType, fieldDefinition: GraphQLFieldDefinition): DataFetcher<Cypher>? {
-            if (rootType.name != MUTATION){
+            if (rootType.name != MUTATION) {
                 return null
             }
             if (fieldDefinition.cypherDirective() != null) {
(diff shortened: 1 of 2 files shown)


al-jeyapal:
@jexp any chance we can get 1 version 1.0.1 published with this fix?

PR #99, opened 2020-08-03
Add support for custom scalars

### LINKED ISSUE #9: support custom scalars
- todo how do we to the mapping / coercion?
- should that be a client side concern
- do we need additional directives or @cypher?
- can we go with some valueOf/toString from the JVM?

see in the javascript library

Andy2003:
This PR add support for custom scalars

For the given schema:

```graphql
scalar Date
type Movie {
  _id: ID!
  title: String!
  released: Date
}
```

you can now run queries like:

```graphql
{
  movie(filter:{released_gte: 1994}) {
    title
    released
  }
}
```

resolves #9

### DIFF
--- a/src/test/resources/translator-tests-custom-scalars.adoc
+++ b/src/test/resources/translator-tests-custom-scalars.adoc
@@ -1,0 +1,202 @@
+:toc:
+
+= Translator Tests
+
+== Schema
+
+[source,graphql,schema=true]
+----
+scalar Date
+type Movie {
+  _id: ID!
+  title: String!
+  released: Date
+}
+----
+
+== Tests
+
+=== Create
+
+.GraphQL-Query
+[source,graphql]
+----
+mutation {
+  createMovie(title:"Forrest Gump", released: 1994) {
+    title
+    released
+  }
+}
+----
+
+.Cypher params
+[source,json]
+----
+{
+  "createMovieTitle": "Forrest Gump",
+  "createMovieReleased": 1994
+}
+----
+
+.Cypher
+[source,cypher]
+----
+CREATE (createMovie:Movie { title: $createMovieTitle, released: $createMovieReleased })
+WITH createMovie
+RETURN createMovie { .title, .released } AS createMovie
+----
+
+=== Update
+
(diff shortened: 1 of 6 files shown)


src/test/resources/translator-tests-custom-scalars.adoc
```diff
@@ -0,0 +1,202 @@
...
+  "updateMovieReleased": null
+}
+----
+
+.Cypher
+[source,cypher]
+----
+MATCH (updateMovie: Movie)
+WHERE ID(updateMovie) = toInteger($updateMovie_id)
+SET updateMovie = { released: $updateMovieReleased }
```
jexp, line 140: this should probably be +=
as it should not override other existing properties.

jexp: Or use`REMOVE updateMovie.released` for null fields
or just use `SET updateMovie.released = $updateMovieReleased`

Andy2003: This has nothing to do with this PR, since its common setter logic, that update is `SET x =` and merge is `SET x +=` If we want to change this behavior, we should create an extra ticket for.

jexp: then that logic is odd, as SET x = {} should only be there for create.

For update (as you might leave off properties) it should always be +=, never =

if that's not the case it's a bug and we need to fix it please create the GH issue.

Andy2003: So what is than the difference between the `merge` and `update`-fields? The SET x = {} was also the problem of #95

jexp: Usually you merge on a single field (in exceptional cases on a set of fields). (they keys that are used to look-up nodes for uniqueness)
And update fields are the fields that are added/overriden.

As MERGE is get-or-create you don't want to destroy existing additional data that might have been added in between by other means.

And if you leave off fields it doesn't mean you want to remove them, just that you don't have data (or updates) for them at the moment.

PR #101, opened 2020-08-05
Prevent use of variables when limit and offset are specified directly as numbers

Andy2003:
This change prevents the use of variables when limit and offset are specified directly as numbers.

Prior this change the graphql query:

```graphql
{ person:person(first:2,offset:3) { age } }
```

resulted in 

```cypher
MATCH (person:Person)
RETURN person { .age } AS person SKIP $personOffset LIMIT $personFirst
```

after this change the following cypher will be generated:

```cypher
MATCH (person:Person)
RETURN person { .age } AS person SKIP 3 LIMIT 2
```

This simplifies testing, as the generated cypher can be used directly in the neo4j browser without the manual step of replacing the variables for limit and offset


jexp:
But it will prevent the use of query plan cache. We should always use parameters

Andy2003:
Good point, I just took a look at the js-lib and there no parameter were generated for this skip and limit

jexp:
Perhaps create a GH issue there then?

PR #103, opened 2020-08-05
Nested sorting on fields

### LINKED ISSUE #3: nested sorting also on fields
Support sorting not just on query fields a but also on other ObjectReferences (Relationships and Nodes), needs to use apoc functions for sorting collections inline.

Andy2003:
This commit add the following features:

* sorting on multiple properties
* sorting on fields
* augmentation-tests provides a diff on failure that can be viewed in IntelliJ

Bugfix:

* The formatted field for temporal is now a string rather than an object

resolves #3


src/main/kotlin/org/neo4j/graphql/SchemaBuilder.kt
```diff
@@ -99,8 +100,8 @@ object SchemaBuilder {
         return handler
     }
 
-    private fun augmentSchema(sourceSchema: GraphQLSchema, handler: List<AugmentationHandler>): GraphQLSchema {
-        val types = sourceSchema.typeMap.toMutableMap()
+    private fun augmentSchema(sourceSchema: GraphQLSchema, handler: List<AugmentationHandler>, config: SchemaConfig): GraphQLSchema {
+        val types = sourceSchema.typeMap.toMap(ConcurrentHashMap())
```
jexp, line 104: why do we need the CHM?

src/main/kotlin/org/neo4j/graphql/handler/QueryHandler.kt
```diff
@@ -34,7 +34,8 @@ class QueryHandler private constructor(
                 .argument(input(OFFSET, Scalars.GraphQLInt))
                 .type(GraphQLNonNull(GraphQLList(GraphQLNonNull(GraphQLTypeReference(type.name)))))
             if (orderingTypeName != null) {
-                builder.argument(input(ORDER_BY, GraphQLTypeReference(orderingTypeName)))
+                val orderType = GraphQLList(GraphQLNonNull(GraphQLTypeReference(orderingTypeName)))
```
jexp, line 37: graphql automatically supports lists instead of single elements here for enums.

Andy2003: Yes but we should define the schema correctly

src/main/kotlin/org/neo4j/graphql/SchemaBuilder.kt
Andy2003: Wen we iterate over the types, the ordering-type may be added so it will throw an concurrent modification exception. This is a simple solution. Otherwise I need to split the logic up to first generate all required types and then enhance the types in a second run.

src/main/kotlin/org/neo4j/graphql/handler/QueryHandler.kt
jexp: won't that break single element sorts then? that already exist?

Andy2003: I tested it with GraphiQL and got no errors. And the tests are also green.
Can you point me to a doc describing the behavior you explained?

Andy2003:
fixes #115

src/main/kotlin/org/neo4j/graphql/SchemaBuilder.kt
jexp: yeah that would be better, can you create an issue so that we don't forget to fix this in the future?

src/main/kotlin/org/neo4j/graphql/handler/projection/ProjectionBase.kt
```diff
@@ -16,21 +16,30 @@ open class ProjectionBase {
...
             }
-        }
-        @Suppress("SimplifiableCallChain")
-        return if (values == null) ""
-        else " ORDER BY " + values
-            .map { it.split("_") }
-            .map { "$variable.${it[0]} ${it[1].toUpperCase()}" }
-            .joinToString(", ")
+            ?.map {
+                val index = it.lastIndexOf('_')
```
jexp, line 38: good one, this also fixes another reported issue #115

src/main/kotlin/org/neo4j/graphql/handler/projection/ProjectionBase.kt
```diff
@@ -206,7 +214,11 @@ open class ProjectionBase {
         } ?: when {
             isObjectField -> {
                 val patternComprehensions = if (fieldDefinition.isNeo4jType()) {
-                    projectNeo4jObjectType(variable, field)
+                    if (neo4jFieldsToPass.contains(fieldDefinition.innerName())) {
```
jexp, line 217: could we give more of a semantic meaning to this expression?

src/main/kotlin/org/neo4j/graphql/handler/projection/ProjectionBase.kt
```diff
@@ -230,7 +242,7 @@ open class ProjectionBase {
             .filterIsInstance<Field>()
             .map {
                 val value = when (it.name) {
-                    NEO4j_FORMATTED_PROPERTY_KEY -> "$variable.${field.name}"
+                    NEO4j_FORMATTED_PROPERTY_KEY -> "toString($variable.${field.name})"
```
jexp, line 245: why did we have to do this?

src/main/kotlin/org/neo4j/graphql/handler/projection/ProjectionBase.kt
```diff
@@ -336,9 +348,27 @@ open class ProjectionBase {
...
+        val orderBy = getOrderByArgs(field.arguments)
+        val sortByNeo4jTypeFields = orderBy
+            .filter { (property, _) -> nodeType.getFieldDefinition(property)?.isNeo4jType() == true }
+            .map { (property, _) -> property }
+            .toSet()
+
+        val fieldProjection = projectFields(childVariable, field, nodeType, env, variableSuffix, sortByNeo4jTypeFields)
+        var comprehension = "[($variable)$inArrow-[$relPattern]-$outArrow($endNodePattern)${where.query} | ${fieldProjection.query}]"
+        if (orderBy.isNotEmpty()) {
+            val sortArgs = orderBy.joinToString(", ", transform = { (property, direction) -> if (direction == Sort.ASC) "'^$property'" else "'$property'" })
```
jexp, line 361: hopefully we can soon handle this with subqueries

src/main/kotlin/org/neo4j/graphql/handler/projection/ProjectionBase.kt
```diff
@@ -336,9 +348,27 @@ open class ProjectionBase {
...
+            comprehension = "apoc.coll.sortMulti($comprehension, [$sortArgs])"
+            if (sortByNeo4jTypeFields.isNotEmpty()) {
+                val neo4jFiledSelection = field.selectionSet.selections
```
jexp, line 364: field not filed

src/main/kotlin/org/neo4j/graphql/handler/projection/ProjectionBase.kt
```diff
@@ -336,9 +348,27 @@ open class ProjectionBase {
...
+                    .filter { selection -> sortByNeo4jTypeFields.contains((selection as? Field)?.name) }
+                val deferredProjection = projectSelection("sortedElement", neo4jFiledSelection, nodeType, env, variableSuffix)
+                    .map { cypher -> cypher.query }
+                    .joinNonEmpty(", ")
+                comprehension = "[sortedElement IN $comprehension | sortedElement { .*, $deferredProjection }]"
```
jexp, line 369: do we really want `.*` ? this can pull way too many properties that are not selected / asked for?
What do we need the deferredProjection for?

jexp: Are those sort fields that are not in the projection?

src/test/resources/movie-tests.adoc
```diff
@@ -857,6 +859,107 @@ RETURN  movie  {  .title  } AS  movie
...
+
+.GraphQL-Query
+[source,graphql]
+----
+{
+  Movie(orderBy:title_desc) {
+    title
+    actors(orderBy:name_desc) {
+      name
+      movies(orderBy:[title_asc, title_desc]) {
```
jexp, line 872: should that be allowed to sort by the same property twice in two ways?

src/test/resources/movie-tests.adoc
```diff
@@ -857,6 +859,107 @@ RETURN  movie  {  .title  } AS  movie
...
+----
+MATCH (user:User)
+RETURN user {
+  .name,
+  born: {
+    formatted: toString(user.born)
+  },
+  knows: [sortedElement IN apoc.coll.sortMulti([(user)-[:KNOWS]->(userKnows:User) | userKnows {
+    .name,
+    born: userKnows.born,
```
jexp, line 943: pretty impressive statement :) with the nested sorted.

src/test/resources/movie-tests.adoc
```diff
@@ -857,6 +859,107 @@ RETURN  movie  {  .title  } AS  movie
...
+    knows: [sortedElement IN apoc.coll.sortMulti([(userKnows)-[:KNOWS]->(userKnowsKnows:User) | userKnowsKnows {
+      .name,
+      born: userKnowsKnows.born
+    }], ['born','born']) | sortedElement {
```
jexp, line 947: should one of these be ascending?

jexp: do we need to differentiate between nested levels?

src/test/resources/movie-tests.adoc
```diff
@@ -857,6 +859,107 @@ RETURN  movie  {  .title  } AS  movie
...
+[source,graphql]
+----
+{
+  User(orderBy: born_desc){
+    name
+    born {formatted}
+    knows(orderBy: born_asc) {
+      name
+      born {formatted}
+      knows(first: 2, offset: 1, orderBy: [born_desc, born_desc]) {
```
jexp, line 914: one of them should be asc?

src/test/resources/movie-tests.adoc
```diff
@@ -996,21 +1098,22 @@ mutation {
...
-MATCH (from:Genre { name: $fromName })
-MATCH (to:Movie) WHERE to.movieId IN $toMovies
-MERGE (from)<-[:IN_GENRE]-(to)
-WITH DISTINCT from AS addGenreMovies
-RETURN addGenreMovies { .name } AS addGenreMovies
+CREATE (createUser:User:Person { userId: $createUserUserId })
+WITH createUser
+RETURN createUser {
+  .userId,
+  __typename: head( [ label IN labels(createUser) WHERE label IN $createUserValidTypes ] )
```
jexp, line 1115: what was this again?

src/test/resources/translator-tests1.adoc
```diff
@@ -34,7 +34,7 @@ interface Location {
 type City implements Location {
   name: String
   founded: Person @relation(name:"FOUNDED", direction: IN)
-  cityArg: String
+  city_arg: String
```
jexp, line 37: probably use `city_Arg` to make sure the casing stays correct too

jexp (approved):
Looks good to me, can you quickly check my comments, then we can merge it.

src/main/kotlin/org/neo4j/graphql/handler/projection/ProjectionBase.kt
Andy2003: This was a Bug I discovered by comparing to the js-lbrary. In case we should return the `formatted` Date/DateTime/* property, this ensures, that a string will be returned instead the Date/DateTime/*-Object.

jexp:
Also needs to be rebased.

src/main/kotlin/org/neo4j/graphql/handler/projection/ProjectionBase.kt
Andy2003: This logic I copied over form the js-library as well. The selection of the *. properties is done in the pattern comprehension so the `sortedElement` will already be _projected_

src/test/resources/movie-tests.adoc
Andy2003: This test is taken over form the js-library. No it does not make any sense. I adjust it.

src/test/resources/movie-tests.adoc
Andy2003: Do you mean to have different names?

src/test/resources/movie-tests.adoc
Andy2003: This is to have the name of the type when using interfaces, so that you know what concrete type the current entry is of.

src/main/kotlin/org/neo4j/graphql/handler/projection/ProjectionBase.kt
```diff
@@ -16,21 +16,30 @@ open class ProjectionBase {
...
-        val values = arg?.value?.let { it ->
-            when (it) {
-                is ArrayValue -> it.values.map { it.toJavaValue().toString() }
-                is EnumValue -> listOf(it.name)
-                is StringValue -> listOf(it.value)
-                else -> null
+        return arg?.value
+            ?.let { it ->
+                when (it) {
+                    is ArrayValue -> it.values.map { it.toJavaValue().toString() }
```
jexp, line 31: don't we need to check here as well Enum/StringValue?

src/main/kotlin/org/neo4j/graphql/handler/projection/ProjectionBase.kt
```diff
@@ -16,21 +16,30 @@ open class ProjectionBase {
     }
 
     fun orderBy(variable: String, args: MutableList<Argument>): String {
+        val values = getOrderByArgs(args)
+        if (values.isEmpty()) {
+            return ""
+        }
+        return " ORDER BY " + values.joinToString(", ", transform = { (property, direction) -> "$variable.$property $direction" })
```
jexp, line 23: do we might have to quote the property? if it has special characters/spaces? e.g. from an alias

jexp (approved):
you can check my comments for a later fix.

src/main/kotlin/org/neo4j/graphql/handler/projection/ProjectionBase.kt
Andy2003: `it.toJavaValue()` does the check

Issue #100, opened 2020-08-04
Symmetrical Relationships?


michael-forman:
Hi,

I'm pretty new to graphQL and cypher, and am having some issues figuring out how to model, and query for symmetrical relationships using neo4j-graphql-java (1.0.0)

I have an example schema of:
```
      type User {
        uuid: ID!
        name: String!
        buddies: [Buddy!] @relation(name:"BUDDY")
      }

      type Buddy @relation(name:"BUDDY") {
        _id: ID!
        from: User!
        since: Int # eg year
        # Potentially other fields
        to: User!
      }
```

(I put the `@relation` directive on User buddies because without it the direction seems to be backwards, I think that is already similar to what is reported in issue 98)

In this case I want the "Buddy" relationship to be symmetrical.  I understand that in neo4j all relationships are directional, however they can be queried without the direction.

With two Users, Person 1 (uuid:1) and Person 2 (uuid: 2) and a Buddy relation from Person 1 to Person 2
we have
![users](https://user-images.githubusercontent.com/12082333/89258758-2132a280-d66c-11ea-877a-4d460e044976.png)

a grapgQL query of `query { user(uuid: "1") { name buddies { _id to { name } }  } }`  generates the expected output, ie it includes the Person 2 buddy relationship.

The cypher is:
```
MATCH (user:User) WHERE user.uuid = "1"
RETURN user { .name, buddies:[(user)-[userBuddies:BUDDY]->(userBuddiesTo:User) | userBuddies { _id:ID(userBuddies), to:userBuddiesTo { .name } }] } AS user
```

Changing to uuid 2 and because of the direction, I don't get to see the buddy.
cypher:
```
MATCH (user:User) WHERE user.uuid = "2"
RETURN user { .name, buddies:[(user)-[userBuddies:BUDDY]->(userBuddiesTo:User) | userBuddies { _id:ID(userBuddies), to:userBuddiesTo { .name } }] } AS user
```

If I remove the direction on the generated cypher
```
MATCH (user:User) WHERE user.uuid = "2"
RETURN user { .name, buddies:[(user)-[userBuddies:BUDDY]-(userBuddiesTo:User) | userBuddies { _id:ID(userBuddies), to:userBuddiesTo { .name } }] } AS user
```

then the buddy relationship is seen.

So:

- Is there a supported mechanism to achieve what I'm trying to do here?
- There will be other "symmetrical" relationships too that may not require a "rich" relation, is that possible?

[...]

PR #104, opened 2020-08-06
Make it clearer the BOTH is an option for the relation direction

michael-forman:
Related to my question in issue #100 I think these minor changes would have enabled me to figured out how to do what I wanted without having to look at the source code.


jexp:
Oh @michael-forman would you mind signing our CLA for your contribution?

https://neo4j.com/developer/cla/

Issue #45, opened 2019-07-28
A schema for tree-like structure fails with duplicate fields


lilianaziolek:
My graph has a tree-like structure, where I have Category nodes, they (optionally) point to their parent category via a relation PARENT_CATEGORY. I am trying to create a GraphQL schema that would allow me to take out a category and its parent and immediate children (if any) and this fails with Schema Problem:

```
SchemaProblem{errors=[The type 'Mutation' [@-1:-1] has declared a field with a non unique name 'addCategoryParentCategory', The type 'Mutation' [@-1:-1] has declared a field with a non unique name 'deleteCategoryParentCategory']}
```

The schema looks like so:
```graphql
type Category {
    url: ID!
    name: String
    order: Int
    parentCategory: Category @relation(name: "PARENT_CATEGORY", direction: OUT)
    subCategories: [Category] @relation(name: "PARENT_CATEGORY", direction: IN)
}
```
Is it possible to get this structure to work in GraphQL? Is there a workaround for this issue, or would it need to be fixed in neo4j-graphql-java?

jexp:
it should work. thanks for raising the issue.

it's probably a bug/oversight in the mutation generation.

lilianaziolek:
As a side note - it does work where you have a separate type for Relation (that is what I submitted the test for in PR). I just didn't need this scenario back then, and now that I do, I discovered this does not work :)

jexp:
ya, the mutation generation is also more recent than your contribution back then. 
it's obviously a case we've overlooked, sorry for that.

lilianaziolek:
I feel like perhaps I should just give you my schema file so that you can bash it on some real-world scenarios 😄

jexp:
That would definitely help. :)

I need to add the ability to embed a schema into a TCK file then it can be self-contained.

Andy2003:
resolved by #107 (#98)

PR #107, opened 2020-08-07
Fix for relationship mutations are not using the correct direction

### LINKED ISSUE #98: Relationship mutations are not using the correct direction
I have a database with nodes like this:

graph

And a graphql schema like this:

```
type Team {
  id: ID!
  name: String!
  players: [Player!]! @relation(name: "MEMBER_OF", direction: IN)
}

type Player {
  id: ID!
  name: String!
  teams: [Team] @relation(name: "MEMBER_OF")
}
```

The cypher queries generated when I run the deleteTeamPlayers and addTeamPlayers mutations are specifying the relationship in the wrong direction.

Here's the cypher generated by deleteTeamPlayers:

> RUN "MATCH (from:Team { id: $fromId }) MATCH (to:Player) WHERE  to.id IN $toPlayers **MATCH (from)-[r:MEMBER_OF]->(to)** DELETE r WITH DISTINCT from AS deleteResult RETURN deleteResult { .name } AS deleteResult"

And here's addTeamPlayers:

> RUN "MATCH (from:Team { id: $fromId }) MATCH (to:Player) WHERE  to.id IN $toPlayers **MERGE (from)-[:MEMBER_OF]->(to)** WITH DISTINCT from AS addResult RETURN addResult { .name } AS addResult"

These queries are building MATCH and MERGE clauses going from Team to Player even though my graphQL schema and database have the relationships going from Player to Team. The result is that the deleteTeamPlayers mutation doesn't do anything since it can't find any relationships to match on, and the addTeamPlayers mutation adds a relationship in the opposite direction of what's desired.

The deletePlayerTeams and addPlayerTeams mutations that were generated from the Player type work fine. It seems like the problem is that these relationship mutations are always using an outbound direction.

Andy2003:
This bugfix solves the problem that the mapping was wrong for incoming relations

(resolves #98)

### DIFF
--- a/src/main/kotlin/org/neo4j/graphql/GraphQLExtensions.kt
+++ b/src/main/kotlin/org/neo4j/graphql/GraphQLExtensions.kt
@@ -49,9 +49,13 @@ fun GraphQLFieldsContainer.relationshipFor(name: String): RelationshipInfo? {
             ?: throw IllegalArgumentException("$name is not defined on ${this.name}")
     val fieldObjectType = field.type.inner() as? GraphQLFieldsContainer ?: return null
 
-    val (relDirective, isRelFromType) = if (isRelationType()) {
+    val (relDirective, inverse) = if (isRelationType()) {
+        val typeName = this.name
         (this as? GraphQLDirectiveContainer)
-            ?.getDirective(DirectiveConstants.RELATION)?.let { it to false }
+            ?.getDirective(DirectiveConstants.RELATION)?.let {
+                // do inverse mapping, if the current type  is the `to` mapping of the relation
+                it to (fieldObjectType.getFieldDefinition(it.getArgument(RELATION_TO, null))?.name == typeName)
+            }
                 ?: throw IllegalStateException("Type ${this.name} needs an @relation directive")
     } else {
         (fieldObjectType as? GraphQLDirectiveContainer)
@@ -60,11 +64,8 @@ fun GraphQLFieldsContainer.relationshipFor(name: String): RelationshipInfo? {
                 ?: throw IllegalStateException("Field $field needs an @relation directive")
     }
 
-    // TODO direction is depending on source/target type
+    val relInfo = relDetails(fieldObjectType, relDirective)
 
-    val relInfo = relDetails(fieldObjectType) { argName, defaultValue -> relDirective.getArgument(argName, defaultValue) }
-
-    val inverse = isRelFromType && fieldObjectType.getFieldDefinition(relInfo.startField)?.name != this.name
     return if (inverse) relInfo.copy(out = relInfo.out?.let { !it }, startField = relInfo.endField, endField = relInfo.startField) else relInfo
 }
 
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/neo4j/graphql/GraphQLExtensions.kt
```diff
@@ -49,9 +49,13 @@ fun GraphQLFieldsContainer.relationshipFor(name: String): RelationshipInfo? {
...
     val fieldObjectType = field.type.inner() as? GraphQLFieldsContainer ?: return null
 
-    val (relDirective, isRelFromType) = if (isRelationType()) {
+    val (relDirective, inverse) = if (isRelationType()) {
+        val typeName = this.name
         (this as? GraphQLDirectiveContainer)
-            ?.getDirective(DirectiveConstants.RELATION)?.let { it to false }
+            ?.getDirective(DirectiveConstants.RELATION)?.let {
+                // do inverse mapping, if the current type  is the `to` mapping of the relation
+                it to (fieldObjectType.getFieldDefinition(it.getArgument(RELATION_TO, null))?.name == typeName)
```
jexp, line 57: we need to be careful, this doesn't always work. if you have a rich relationship which has the same type in to/from then we need to use the explicit information from that type.

jexp: we need the information the rel-field then I mean that indicates the correct direction otherwise it's "random" if from/to is matching the type name.

src/main/kotlin/org/neo4j/graphql/GraphQLExtensions.kt
```diff
@@ -49,9 +49,13 @@ fun GraphQLFieldsContainer.relationshipFor(name: String): RelationshipInfo? {
             ?: throw IllegalArgumentException("$name is not defined on ${this.name}")
     val fieldObjectType = field.type.inner() as? GraphQLFieldsContainer ?: return null
 
-    val (relDirective, isRelFromType) = if (isRelationType()) {
+    val (relDirective, inverse) = if (isRelationType()) {
```
jexp, line 52: perhaps we can simplify this code a bit so it gets easier to understand.

src/main/kotlin/org/neo4j/graphql/GraphQLExtensions.kt
jexp: see: https://github.com/neo4j-graphql/neo4j-graphql-java/issues/45

src/main/kotlin/org/neo4j/graphql/GraphQLExtensions.kt
```diff
@@ -60,11 +64,8 @@ fun GraphQLFieldsContainer.relationshipFor(name: String): RelationshipInfo? {
                 ?: throw IllegalStateException("Field $field needs an @relation directive")
     }
 
-    // TODO direction is depending on source/target type
-
-    val relInfo = relDetails(fieldObjectType) { argName, defaultValue -> relDirective.getArgument(argName, defaultValue) }
+    val relInfo = relDetails(fieldObjectType, relDirective)
 
-    val inverse = isRelFromType && fieldObjectType.getFieldDefinition(relInfo.startField)?.name != this.name
```
jexp, line 67: where is this bit handled now?

jexp: this was the fix for issue #45

Andy2003: I moved it to https://github.com/Andy2003/neo4j-graphql-java/blob/bugfix/%2398-fix-relationship-mapping/src/main/kotlin/org/neo4j/graphql/GraphQLExtensions.kt#L57 since it is only relevant for rich relations

src/test/resources/relationship-tests.adoc
```diff
@@ -0,0 +1,323 @@
+:toc:
+
+= Movie Test TCK
```
greg-katz, line 3: Should this be `Relationship Test TCK` rather than `Movie Test TCK`?

jexp: Thanks for spotting this, fixed in master.

PR #117, opened 2020-08-28
This change fixes an issue where properties got unset for when updating a node

### LINKED ISSUE #105: Don't use assignment for MERGE/Update
See the discussion here:

https://github.com/neo4j-graphql/neo4j-graphql-java/pull/99#pullrequestreview-461213019

Switch from `SET x = {...}` to `SET x += {}` or individual set operations.
We could also consider putting the data just into a parameter instead. So it would become `SET x += $data`

> Usually you merge on a single field (in exceptional cases on a set of fields). (they keys that are used to look-up nodes for uniqueness)
> And update fields are the fields that are added/overriden.

> As MERGE is get-or-create you don't want to destroy existing additional data that might have been added in between by other means.

> And if you leave off fields it doesn't mean you want to remove them, just that you don't have data (or updates) for them at the moment.

Andy2003:
Now we always use `SET x += {}` to update properties.

fixes #105

### DIFF
--- a/src/main/kotlin/org/neo4j/graphql/handler/MergeOrUpdateHandler.kt
+++ b/src/main/kotlin/org/neo4j/graphql/handler/MergeOrUpdateHandler.kt
@@ -67,13 +67,8 @@ class MergeOrUpdateHandler private constructor(
     }
 
     init {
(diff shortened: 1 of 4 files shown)


jexp (approved):
Looks good, thanks a lot!

PR #120, opened 2020-08-29
update graphql-java to version 15

claymccoy:
This is an old version of a core dependency used in most graphql related projects in java. If this dependency isn't updated then you have conflicts with incompatible transitive versions of graphql-java.
Unfortunately, it looks like a committer to graphql-java likes to make backwards a lot of uneccessarily incompatible changes to the library. So I had to make a lot of changes for this to compile and for the tests to pass. I had to do some archeology in the history of graphql-java to understand some of the changes.

* GraphQLType.name no longer exists. simplePrint(GraphQLType) is used instead
* GraphQLFieldDefinition is no longer a GraphQLType, but it looks like GraphQLFieldDefinition.originalType will get the type we are looking for.
* ScalarInfo.STANDARD_SCALAR_DEFINITIONS was replaced with ScalarInfo.GRAPHQL_SPECIFICATION_SCALARS_DEFINITIONS
* The value generic for the return type of GraphQLSchema.typeMap is now more specific so BuildingEnv's constructor had to change.
* ShemaPrinter lost some include methods. One was previously mispelled.
* DefaultSchemaPrinterComparatorRegistry became DefaultGraphqlTypeComparatorRegistry.

I had to make some changes to the tests to make them pass, which makes me quite uncomfortable with this. Please check these over to see if I should have made more changes. I'm going ahead with this because the overall functionality still seems legit. I have no idea why these were needed.
* In translator-tests1 'render values' I had to move the clauses of the where clause around. It mateches the order of the cypher params now, so it kind of makes sense.
* The one line GraphQL query no longer works. It had to be broken up to multilines in several tests.
* relationship-tests is a weird one. I changed the generated Cypher, but it makes more sense now. The relationship is player through memberships to team. So `playerMembershipsTeam: Team` makes a lot more sense than the original `playerMembershipsPlayer: Player`. It actually looks like it was incorrect originally.


jexp:
@Andy2003 please help with the review and rebase due to the test changes.

Andy2003:
@claymccoy Thank you for your contribution

I walked through the changes and adjusted the branch in (https://github.com/claymccoy/neo4j-graphql-java/pull/1)

I replaced the massive use of `simplePrint` by a fallback `name()` method 

> DefaultSchemaPrinterComparatorRegistry became DefaultGraphqlTypeComparatorRegistry.

This was only used to sort the fields of the type for better comparison / diffing. This is no longer possible so I romed the whole block

> The one line GraphQL query no longer works. It had to be broken up to multilines in several tests.

There were `nbsp` characters in the tests. replacing them by space make them run again

> In translator-tests1 'render values' I had to move the clauses of the where clause around. It mateches the order of the cypher params now, so it kind of makes sense.

I got the same issue with this test in the past. The test is OK so far.

> relationship-tests is a weird one. I changed the generated Cypher, but it makes more sense now. The relationship is player through memberships to team. So playerMembershipsTeam: Team makes a lot more sense than the original playerMembershipsPlayer: Player. It actually looks like it was incorrect originally.

This was indeed a bug

claymccoy:
@Andy2003 I pulled your changes in. Thanks!
Is it possible to make a release once these changes are in? I'd really like to use it in a project.

pom.xml
```diff
@@ -88,7 +88,7 @@
         <dependency>
             <groupId>com.graphql-java</groupId>
             <artifactId>graphql-java</artifactId>
-            <version>12.0</version>
+            <version>15.0</version>
```
Andy2003, line 91: @jexp Do you think we need to add a compatibility layer to support older versions of `graphql-java`. Since they are not very compatible code my break. E.g. I'm using [graphql-spqr-spring-boot-starter](https://github.com/leangen/graphql-spqr-spring-boot-starter) with this library, which has not yet been updated to use version 15.0 of `graphql-java`. There is an open issue for this https://github.com/leangen/graphql-spqr/issues/356 . So also other users my have this compatibility issues.

src/main/kotlin/org/neo4j/graphql/BuildingEnv.kt
```diff
@@ -23,7 +25,7 @@ class BuildingEnv(val types: MutableMap<String, GraphQLType>) {
             type = GraphQLNonNull(type)
         }
         return GraphQLFieldDefinition.newFieldDefinition()
-            .name("$prefix${resultType.name}")
+            .name("$prefix${simplePrint(resultType)}")
```
Andy2003, line 28: use `resultType.name()`

src/main/kotlin/org/neo4j/graphql/BuildingEnv.kt
```diff
@@ -71,7 +73,7 @@ class BuildingEnv(val types: MutableMap<String, GraphQLType>) {
         }
         val existingFilterType = types[filterName]
         if (existingFilterType != null) {
-            return (existingFilterType as? GraphQLInputType)?.name
+            return simplePrint(existingFilterType as? GraphQLInputType)
```
Andy2003, line 76: use `(existingFilterType as? GraphQLInputType)?.name()`

src/test/kotlin/org/neo4j/graphql/TranslatorExceptionTests.kt
```diff
@@ -16,16 +16,26 @@ class TranslatorExceptionTests : AsciiDocTestSuite("translator-tests1.adoc") {
     }
 
     override fun schemaTestFactory(schema: String): List<DynamicNode> {
-        val translator = Translator(SchemaBuilder.buildSchema(schema));
+        val translator = Translator(SchemaBuilder.buildSchema(schema))
         return listOf(
                 DynamicTest.dynamicTest("unknownType") {
                     Assertions.assertThrows(IllegalArgumentException::class.java) {
-                        translator.translate(" { company { name } } ")
+                        translator.translate("""
```
Andy2003, line 23: these changes can be reverted, since there was no issue with singel line graphql but with nbsp-characters

Andy2003 (changes requested):

Andy2003:
> Is it possible to make a release once these changes are in? I'd really like to use it in a project.

@jexp is the master of releases ;-)

jexp:
Let me review and merge this in and then the few open PRs from @Andy2003 and then I promise to do the release this week.

Oh btw. @claymccoy can you please sign our CLA first? https://neo4j.com/developer/cla

claymccoy:
> Oh btw. @claymccoy can you please sign our CLA first? https://neo4j.com/developer/cla

email sent for cla

pom.xml
claymccoy: That was similar to the position I was in and why I tried to help update the library here. I wanted to use a library similar to spqr (graphql-kotlin) which uses a later graphql-java. Rather than a compatibility layer, I opted to help move this project forward with the thinking that a rising tide lifts all boats. I could help update other projects too. This is a common problem with libraries that use common dependencies. It is often an issue with Guava or something similar. The two strategies I've seen most often used successfully are to severely limit dependencies used in a library and/or to be aggressive about getting everyone to update the common deps.

jexp: My suggestion is:
1. we release a version of 1.0.1 with the nested sorting PR merged and graphql-java 12 (and create a branch/tag for that)
2. then we release 1.1.0 with the upgrade of graphql-java to 15 (master)

Andy2003:
@claymccoy please merge https://github.com/claymccoy/neo4j-graphql-java/pull/2 so we can merge this PR afterwards

pom.xml
Andy2003: :+1:

PR #114, opened 2020-08-27
Improve test navigation

Andy2003:
This change adds support for:
* nesting tests (deep hierarchy)
* direct navigation to failed block (cypher or cypher params)
* let augmentation-tests provides a diff on failure that can be viewed in IntelliJ

PR #121, opened 2020-08-31
Add ability to print out additional debug information for cypher-tests

Andy2003:
AS requested by https://github.com/neo4j-graphql/neo4j-graphql-java/pull/114#issuecomment-683479125

### DIFF
--- a/src/test/kotlin/org/neo4j/graphql/utils/CypherTestSuite.kt
+++ b/src/test/kotlin/org/neo4j/graphql/utils/CypherTestSuite.kt
@@ -104,6 +104,6 @@ class CypherTestSuite(fileName: String) : AsciiDocTestSuite(fileName) {
     }
 
     companion object {
-        const val DEBUG = false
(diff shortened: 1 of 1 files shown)


src/test/kotlin/org/neo4j/graphql/utils/CypherTestSuite.kt
```diff
@@ -36,7 +36,7 @@ class CypherTestSuite(fileName: String) : AsciiDocTestSuite(fileName) {
         val result = createTransformationTask(title, schema, codeBlocks)
 
         val tests = mutableListOf<DynamicNode>()
-        if (DEBUG) {
+        if (System.getProperty("debug", "false") == "true") {
```
jexp, line 39: Probably more specific name? Like `neo4j-graphql-java.debug` Otherwise it might conflict with others?

jexp: I also meant to set it in the companion object so that it doesn't have to read it time and again, but it's probably not a biggie.

PR #130, opened 2020-10-19
update sample groovy server

Kubera2017:
Updated sample groovy server with basic logger and Neo4j 4.x diver


jexp:
Thanks a lot!

PR #134, opened 2020-10-21
Add ability to use a variable as filter.

### LINKED ISSUE #133: Using a variable as filter does not work
The following test is failing

GraphQL-Query:

```graphql
query filterQuery($filter: _PersonFilter) { person(filter: $filter) { name }}
```

Query variables

```json
{
  "filter": {
    "name": "Jane",
    "company": {
      "name_ends_with": "ME"
    }
  }
}
```

Andy2003:
resolves #133

### DIFF
--- a/src/main/kotlin/org/neo4j/graphql/handler/QueryHandler.kt
+++ b/src/main/kotlin/org/neo4j/graphql/handler/QueryHandler.kt
@@ -97,7 +97,7 @@ class QueryHandler private constructor(
         } else {
             "($variable:${label()})"
         }
-        val where = where(variable, fieldDefinition, type, propertyArguments(field), field)
+        val where = where(variable, fieldDefinition, type, propertyArguments(field), field, env)
         return Cypher(
                 """MATCH $select${where.query}
                   |RETURN ${mapProjection.query} AS $variable$ordering${skipLimit.query}""".trimMargin(),
--- a/src/main/kotlin/org/neo4j/graphql/handler/QueryHandler.kt
+++ b/src/main/kotlin/org/neo4j/graphql/handler/QueryHandler.kt
@@ -97,7 +97,7 @@ class QueryHandler private constructor(
         } else {
             "($variable:${label()})"
         }
-        val where = where(variable, fieldDefinition, type, propertyArguments(field), field, env)
+        val where = where(variable, fieldDefinition, type, propertyArguments(field), field, env.variables)
         return Cypher(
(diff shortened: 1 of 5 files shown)


Andy2003:
@jexp we need to merge `1.0`-branch to `master` after this PR is merged

src/main/kotlin/org/neo4j/graphql/handler/QueryHandler.kt
```diff
@@ -97,7 +97,7 @@ class QueryHandler private constructor(
         } else {
             "($variable:${label()})"
         }
-        val where = where(variable, fieldDefinition, type, propertyArguments(field), field)
+        val where = where(variable, fieldDefinition, type, propertyArguments(field), field, env)
```
jexp, line 100: I don't think we should pass `env` through to an inner method but just the parameters/variables

src/main/kotlin/org/neo4j/graphql/handler/projection/ProjectionBase.kt
```diff
@@ -99,11 +105,16 @@ open class ProjectionBase {
             .toMap()
     }
 
-    private fun filterExpressions(value: Any?, type: GraphQLFieldsContainer): List<Predicate> {
-        // todo variable/parameter
-        return if (value is Map<*, *>) {
-            CompoundPredicate(value.map { (k, v) -> toExpression(k.toString(), v, type) }, "AND").parts
-        } else emptyList()
+    private fun filterExpressions(value: Any?, type: GraphQLFieldsContainer, variables: MutableMap<String, Any>): List<Predicate> {
```
jexp, line 108: can we use Map<> here not MutableMap

src/test/resources/filter-tests.adoc
```diff
@@ -1837,37 +1837,39 @@ RETURN graphql.labels(company) AS _labels,
 
 === Filter as variable
 
-CAUTION: Is not working
-
 .GraphQL-Query
 [source,graphql]
```
jexp, line 1841: do we also have a test where inside of a filter a variable/parameter is used?

jexp (changes requested):
see my comments

src/test/resources/filter-tests.adoc
Andy2003: Do you mean the `Variable pass through`-Test?

jexp (approved):

PR #139, opened 2020-10-23
Make the capitalization of the Query-Type fields configurable

### LINKED ISSUE #84: make capitalization for top-level queries configurable

Andy2003:
resolves #84


core/src/test/resources/schema-operations-tests.adoc
```diff
@@ -281,3 +281,268 @@ directive @property(name:String) on FIELD_DEFINITION
...
+type Mutation {
+  createLocation(name: String): Location!
+  createPerson(name: String): Person!
+}
+
+type Person {
+  name: String
+}
+
+type _Neo4jDate {
```
jexp, line 322: could we just out-source that and use an `include::neo4j-types.adoc[]` here (and elsewhere where we check the augmented schema)?

PR #148, opened 2020-11-10
fix filtering on directive in combination with filtering on type field

### LINKED ISSUE #147: Filtering on directive in combination with filtering on type field  is broken
given:

```graphql
interface Person {
    name: ID!
    born: Int
    movies: [Movie]
    score(value:Int!): Int @cypher(statement:"RETURN $value")
}

type Actor implements Person {
    name: ID!
    born: Int
    movies: [Movie] @relation(name:"ACTED_IN")
    score(value:Int!): Int @cypher(statement:"RETURN $value")

    namedColleagues(name: String!): [Person] @cypher(statement: "WITH $this AS this MATCH (this)-[:ACTED_IN]->()<-[:ACTED_IN]-(other) WHERE other.name CONTAINS $name RETURN other")
}

type Movie  {
    title: ID!
    released: Int
    tagline: String
    genre: Genre
    actors: [Actor] @relation(name:"ACTED_IN",direction: IN)
}

enum Genre {
    Action, Drama, Family, Horror, SciFi
}
```

and query:

```graphql
query {
    person(name: "Kevin Bacon") {
        born
        ... on Actor {
            namedColleagues(name: "Meg") {
                ... name
            }
        }
        score(value:7)
    }
}
fragment name on Actor { name }
```

the generated  cypher does not distinguish between the different `name` field values:

```cypher
MATCH (person: Person)
WHERE person.name = $personName
RETURN person {
  .born,
  namedColleagues: [personNamedColleagues IN apoc.cypher.runFirstColumnMany('WITH $this AS this, $name AS name WITH $this AS this MATCH (this)-[: ACTED_IN]->()<-[: ACTED_IN]-(other) WHERE other.name CONTAINS $name RETURN other', { this: person, name: $personName }) | personNamedColleagues {
    .name
  }],
  score: apoc.cypher.runFirstColumnSingle('WITH $this AS this, $value AS value RETURN $value', { this: person, value: $personValue })
} AS Person
```

Andy2003:
resolves #147

### DIFF
--- a/core/src/main/kotlin/org/neo4j/graphql/handler/projection/ProjectionBase.kt
+++ b/core/src/main/kotlin/org/neo4j/graphql/handler/projection/ProjectionBase.kt
@@ -215,7 +215,7 @@ open class ProjectionBase {
         val cypherDirective = fieldDefinition.cypherDirective()
         val isObjectField = fieldDefinition.type.inner() is GraphQLFieldsContainer
         return cypherDirective?.let {
-            val directive = cypherDirective(variable, fieldDefinition, field, it, listOf(Translator.CypherArgument("this", "this", variable)))
+            val directive = cypherDirective(variable + field.aliasOrName().capitalize(), fieldDefinition, field, it, listOf(Translator.CypherArgument("this", "this", variable)))
             if (isObjectField) {
                 val patternComprehensions = projectListComprehension(variable, field, fieldDefinition, env, directive, variableSuffix)
                 Cypher(field.aliasOrName() + ":" + patternComprehensions.query, patternComprehensions.params)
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/org/neo4j/graphql/handler/projection/ProjectionBase.kt
```diff
@@ -215,7 +215,7 @@ open class ProjectionBase {
         val cypherDirective = fieldDefinition.cypherDirective()
         val isObjectField = fieldDefinition.type.inner() is GraphQLFieldsContainer
         return cypherDirective?.let {
-            val directive = cypherDirective(variable, fieldDefinition, field, it, listOf(Translator.CypherArgument("this", "this", variable)))
+            val directive = cypherDirective(variable + field.aliasOrName().capitalize(), fieldDefinition, field, it, listOf(Translator.CypherArgument("this", "this", variable)))
```
jexp, line 218: do we have the same problem in other places? 
sounds like a good idea to encapsulate this variable-contextualization into a method on field or so:

field.contextualize(variable) or field.for(variable) ? perhaps you can do that in a separate PR

jexp (approved):

PR #162, opened 2020-12-24
allow regex comparison operator. Included test cases.

Yashesh123:
Hi,

This is my first PR to this repository.

This PR will add support for regular expression based filtering. For that I have added a field operator using "matches" provided by neo4j-cypher-dsl. Which will convert to its equivalent cypher regexp operator (~=).

So for this schema:

enum Gender { female, male }
type Person {
  id : ID!
  name: String
  age: Int
  height: Float
  fun: Boolean
  gender: Gender
  company: Company @relation(name:"WORKS_AT")
  location: _Neo4jPoint
}
type Company {
  _id: ID
  name: String
  employees: [Person] @relation(name:"WORKS_AT", direction: IN)
}
type Query {
  person : [Person]
}

This graphql query example will work:

{ 
	person(filter: { name_matches: "Ja.*" }) { 
		name 
	}
}

This can be example output:

{
  "person" : [ {
    "name" : "Jane"
  } ]
}

This can also solves requirement for case-insensitiveness as cypher queries are already supporting it using a same (~=) regexp operator.


core/src/main/kotlin/org/neo4j/graphql/Predicates.kt
```diff
@@ -205,6 +205,8 @@ enum class FieldOperator(
     C("_contains", "CONTAINS", { lhs, rhs -> lhs.contains(rhs) }),
     SW("_starts_with", "STARTS WITH", { lhs, rhs -> lhs.startsWith(rhs) }),
     EW("_ends_with", "ENDS WITH", { lhs, rhs -> lhs.endsWith(rhs) }),
+    MA("_matches", "=~", {lhs, rhs -> lhs.matches(rhs) }),
```
Andy2003, line 208: I would suggest to call it `MATCHES`, since its only usage is for the Cypher parameters. This would make the code more readable.

Andy2003 (changes requested):

core/src/main/kotlin/org/neo4j/graphql/Predicates.kt
Yashesh123: Done. Thanks, @Andy2003

Andy2003 (approved):

Issue #75, opened 2020-04-09
Non-matching filters should return empty lists if required by schema


DirkMahler:
The following schema defines a query taking a filter as input object and returning a non-nullable list:

```
type Query {
  person(filter: _PersonFilter) : [Person]!
 }
```

Currently a filter that does not match any person returns nothing (i.e. no key) in the result for that list which according to the schema is invalid.

PR #172, opened 2021-02-25
Added alias or field name as variable to Cypher.kt

DirkMahler:
The alias or field name has been added as variable to the Cypher class. This comes for two purposes:

* The "guessing" of the alias/field name based on the first row of the result can be replaced
* It allows non-null checks to solve #75 (see example below)

In a controller the following code can be used:

```
    private Map<String, Object> getData(List<Cypher> translated) {
        Map<String, Object> data = new HashMap<>();
        for (Cypher cypher : translated) {
            String variable = cypher.getVariable();
            GraphQLType type = cypher.getType();
            Result result = session.run(cypher.getQuery(), cypher.getParams());
            if (result.hasNext()) {
                Object queryResult = isListType(type) ? toIterableResult(result, variable) : toSingleResult(result, variable);
                data.put(variable, queryResult);
            } else {
                if ((type instanceof GraphQLNonNull)) {
                    if (isListType(type)) {
                        data.put(variable, emptyList());
                    } else {
                        throw new IllegalStateException("No result available for non-null alias/field " + variable + ".");
                    }
                } else {
                    data.put(variable, null);
                }
            }
        }
        return data;
    }

    private Iterable<?> toIterableResult(Result result, String key) {
        return () -> {
            Iterator<Record> iterator = result.stream().iterator();
            return new Iterator<>() {
                @Override
                public boolean hasNext() {
                    return iterator.hasNext();
                }

                @Override
                public Object next() {
                    return iterator.next().asMap().get(key);
                }
            };
        };
    }

    private Object toSingleResult(Result result, String key) {
        return result.single().asMap().get(key);
    }

    private boolean isListType(GraphQLType type) {
        if (type instanceof GraphQLList) {
            return true;
        } else if (type instanceof GraphQLNonNull) {
            return isListType(((GraphQLNonNull) type).getWrappedType());
        } else {
            return false;
        }
    }
```
PS: I did not update the examples as I felt a bit unsafe about Kotlin (still stuck in the "pure" Java universe). Any guidance appreciated for providing this as well.


Andy2003:
@DirkMahler please take a look at the PR https://github.com/DirkMahler/neo4j-graphql-java/pull/1

Andy2003 (changes requested):
Apply changes from https://github.com/DirkMahler/neo4j-graphql-java/pull/1

Andy2003 (approved):

jexp:
What I don't understand is the unrelated change of the "Cypher" type as result/parameter to String
that makes it inherently less communicating and less safe.

In places where we know we expect Cypher, we should also make use of that.

What does that "variable" actually mean? 
I didn't understand it from the code changes.

Is that one of the column names that the contained cypher statement returns? Do we always know these?

DirkMahler:
The generated Cypher query per row returns a map with exactly one column. This one is the query field or alias that was part of the GraphQL query. To fill the data field of the GraphQL response the name of this field or alias needs to be extracted, until now this was only possible by looking at the first row of the result. Even worse: if the result was empty it was not possible to infer that name, e.g. for returning an empty list if the schema requires a Non-Null value. This now has been made available via `variable`, maybe the name can be subject to discussion.

Andy2003:
@jexp 

> What I don't understand is the unrelated change of the "Cypher" type as result/parameter to String
that makes it inherently less communicating and less safe.

The use of the `Cypher` type in the generation classes was replaced by the Cypher-DSL. AS such, we no longer need this type internally. It is now only the API for this library.

Because the `variable` field is mandatory in the `Cypher` class, we would have needed to initialize this when using it internally. 
At the end I just removed one wrapping of the directives' `statement`-argument, and passed it directly to the generator without wrapping it in with the `cypher` class

PR #175, opened 2021-03-03
allow projection starting query from rich relationship

### LINKED ISSUE #170: wrong mapping starting query from rich relationship
given

Schema:

```graphql
type Movie {
  title: String
  ratings: [Rated] @relation(name:"RATED")
}
interface Person {
  name: String
}
type User {
  name: String
  rated(rating: Int): [Rated]
}
type Rated @relation(name:"RATED", from: "user", to: "movie") {
  user: User
  rating: Int
  movie: Movie
}
```

Query:

```qraphql
query {
    r: rated( rating_gte : 3) {
      rating
      movie {
        title
      }
    }
}
```

I get following cypher:

```cypher
MATCH ()-[r:RATED]->()
WHERE r.rating >= $r_rating 
RETURN r {
	.rating, 
	movie: r {
		.title
	}
} AS r
```

But I would expect something like:

```cypher
MATCH ()-[r:RATED]->()
WHERE r.rating >= $r_rating
RETURN r {
	.rating,
	movie: head([()-[r]->(movie:Movie) | movie {
		.title
	}])
} AS r
```

Andy2003:
resolves #170


jexp:
As long as we're aware that with any <4.3 database without a rel-type-property index that's an super expensive and slow query.

jexp (approved):

PR #178, opened 2021-03-04
Support filtering of null values also via variables

### LINKED ISSUE #47: lack of filtering for null values
It does not seem to be possible to filter using a null value for a property. For example, the following test fails if pasted into FilterTest:

```kotlin
    @Test
    fun testFilterNull(){
        val query = """{ company (name: null) {employees {gender}}}"""
        val expected = """MATCH (company:Company) WHERE company.name IS NULL RETURN company { employees:[(company)<-[:WORKS_AT]-(companyEmployees:Person) | companyEmployees { .gender }] } AS company"""

        assertQuery(schema, query, expected)
    }
```

The generated query contains `company.name = $companyName` instead of `company.name IS NULL`. This does not work in Neo4j if the value is null. Is there alternative syntax that should be used for such query, or is this a bug?

Andy2003:
resolves #47


jexp:
please rebase

PR #195, opened 2021-03-27
Add support for the passThrough argument of the cypher directive.

### LINKED ISSUE #190: Please add passThrough=true support!
We are using Neo4j-GraphQL plugin in production. And we  are in the process of migrating to Neo4j-GraphQL-Java. Our schema uses a lot of "passThrough=true" parameters to return embedded map/objects. Please add this, thanks!

Andy2003:
The new argument `passThrough` of the `@cypher` directive allows the user to return an arbitrary result for the directive

resolves #190

### DIFF
--- a/core/src/main/kotlin/org/neo4j/graphql/DirectiveConstants.kt
+++ b/core/src/main/kotlin/org/neo4j/graphql/DirectiveConstants.kt
@@ -12,6 +12,7 @@ class DirectiveConstants {
 
         const val CYPHER = "cypher"
         const val CYPHER_STATEMENT = "statement"
+        const val CYPHER_PASS_THROUGH = "passThrough"
 
         const val PROPERTY = "property"
         const val PROPERTY_NAME = "name"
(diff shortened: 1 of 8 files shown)


n2iw:
Thanks @Andy2003! passThrough is mostly working now! There is only one small thing, there is no "projection" based on request for the embedded Map/Object. For example, one of my embedded object is called "files" it has 10 properties in Neo4j, but the request only asked one of them, the result is we are still returning all 10 properties for "files". It was processed correctly in Neo4j-GraphQL plugin. Can we have that capability added? Thanks a lot!

Andy2003:
@n2iw 

I guess you are using the `Translator` directly instead of letting graphql resolve the data via the `dataFetchingInterceptor`. If you call your schema as demonstrated in https://github.com/neo4j-graphql/neo4j-graphql-java/blob/0f89393884914e906de9d78c1e432b5a09a2ce53/core/src/test/kotlin/DataFetcherInterceptorDemo.kt you should get the projected fields returned

n2iw:
@Andy2003 

Yes, we are using the translator directly at the moment. We'll change our code to use DataFetchInterceptor as you suggested. I think this PR should be working fine for us. Thank you so much for solving this issue for us, we really appreciate it! Hope this PR gets merged and released soon!

PR #198, opened 2021-03-28
Separate sorting and paging into own input type

### LINKED ISSUE #196: Separate sorting and paging into own input type
For the following input:

```graphql
input NodeInput {
    id: ID
}
```

this code should be generated

```graphql
enum SortDirection {
  """Sort by field values in ascending order."""
  ASC
  """Sort by field values in descending order."""
  DESC
}

input MovieCreateInput {
  id: ID
}

input MovieOptions {
  """Specify one or more MovieSort objects to sort Movies by. The sorts will be applied in the order in which they are arranged in the array."""
sort: [MovieSort]
  limit: Int
  skip: Int
}

"""Fields to sort Movies by. The order in which sorts are applied is not guaranteed when specifying many fields in one MovieSort object."""
input MovieSort {
  id: SortDirection
}

type Query {
  movies(id: ID, options: MovieOptions): [Movie]!
}

//... all the rest

```

Andy2003:
This feature is part of migrating our API to the one generated by `@neo4j/graphql`.
Here we move the fields for paging and sorting into a options input type, which can be passed to queries and projected relational fields

resolves #196

### DIFF
--- a/core/src/main/kotlin/org/neo4j/graphql/BuildingEnv.kt
+++ b/core/src/main/kotlin/org/neo4j/graphql/BuildingEnv.kt
@@ -2,6 +2,7 @@ package org.neo4j.graphql
 
 import graphql.Scalars
 import graphql.schema.*
+import org.neo4j.graphql.handler.projection.ProjectionBase
 
 class BuildingEnv(
         val types: MutableMap<String, GraphQLNamedType>,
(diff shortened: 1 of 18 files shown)


jexp (approved):
Looks good, thanks a lot, feel free to merge.

PR #222, opened 2021-05-07
Make use of sub-queries, to get rid of most apoc calls.

### LINKED ISSUE #214: Try to replace apoc calls by subqueries

Andy2003:
resolves #214

### DIFF
--- a/core/src/test/resources/custom-fields.adoc
+++ b/core/src/test/resources/custom-fields.adoc
@@ -176,12 +176,8 @@ CALL {
 }
 CALL {
 	WITH person
-	CALL {
-		WITH person
-		WITH person AS this
-		RETURN toLower(this.name) AS personLabel LIMIT 1
-	}
-	RETURN personLabel AS personLabel
+	WITH person AS this
+	RETURN toLower(this.name) AS personLabel LIMIT 1
 }
 RETURN person {
 	.name,
@@ -225,12 +221,8 @@ CALL {
 }
 CALL {
 	WITH person
-	CALL {
-		WITH person
-		WITH person AS this
-		RETURN null AS personNullable LIMIT 1
-	}
-	RETURN personNullable AS personNullable
+	WITH person AS this
+	RETURN null AS personNullable LIMIT 1
 }
 RETURN person {
 	.name,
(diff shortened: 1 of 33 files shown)


Andy2003:
only thing left is updating to ne next release of the neo4j-cypher-dsl

michael-simons:
> only thing left is updating to ne next release of the neo4j-cypher-dsl

Feel free to ping me if needed. 
Next week after hack/commit/push is totally fine.

core/src/test/resources/custom-fields.adoc
```diff
@@ -86,10 +86,10 @@ mutation { createPerson(name:"Jill") {name} }
 .Cypher
 [source,cypher]
 ----
-CALL apoc.cypher.doIt('WITH $name AS name CREATE (p:Person { name: \"Test\"+$name }) RETURN p',  {
-	name: $createPersonName
-}) YIELD value
-WITH value[head(keys(value))] AS createPerson
+CALL {
+	WITH $createPersonName AS name
```
jexp, line 90: we need to watch out here for naming conflicts if we rename to a single/simple name, we should still be able to use parameters in the subqueries

core/src/test/resources/custom-fields.adoc
```diff
@@ -169,14 +170,22 @@ query { Person(name:"Jane") {name, label} }
...
 [source,cypher]
 ----
-UNWIND apoc.cypher.runFirstColumnSingle('WITH $name AS name MATCH (p:Person { name: $name }) RETURN p',  {
-	name: $personName
-}) AS person
+CALL {
+	WITH $personName AS name
+	MATCH (p:Person { name: name }) RETURN p AS person LIMIT 1
+}
+CALL {
```
jexp, line 177: why did we have the 2 subqueries here if the 2nd outer just is a pass-through, can we detect that and remove/skip it if it has no actual operations?

core/src/test/resources/custom-fields.adoc
Andy2003: I see no potential conflict here:

- `$createPersonName` is defined by the library
- `name` is the name of the fields argument defined by the user

jexp (approved):
Looks great, thanks so much. Sorry it took so long.

PR #228, opened 2021-05-20
Add a maven plugin to generate the augmented schema

Andy2003:
This will allow the use of a type save graphql API.


jexp (approved):
Looks good, great idea.

Issue #223, opened 2021-05-14
Graphql query to Fetch data between dates(_Neo4jLocalDateTime),


jithuprabhakaran:
How can we write a Graphql query to fetch data between 2 different dates.? Date property that we use has datatype _Neo4jLocalDateTime.

rbramley:
This is related to #93 - the response states that "_This type of filtering for temporal types is not yet implemented and there is currently no workaround to get it working._"

However, it is possible to workaround it (with an obvious drawback around field selection) by declaring a custom `@cypher` query in your schema e.g. 
```
movieDateFiltered(after: _Neo4jDateInput, before: _Neo4jDateInput): [Movie] @cypher(
statement: "MATCH (m:Movie) WHERE m.pub_date >= date(after) AND m.pub_date <= date(before) RETURN { title: m.title, pub_date: m.pub_date }")
```

And then query as:
```
query { 
  movieDateFiltered(
    after: {year:2016, month:12, day:31},
    before: {year:2020, month:1, day:1}
  )
  {
    title
    pub_date { formatted }
  }
}
```

I've used `_Neo4jDate` for simplicity, but you should be able to extend the same approach for datetimes.

jithuprabhakaran:
Hi Robin Bramley,

Thanks for the response. Let me try like this

Thanks

rbramley:
Hi @jithuprabhakaran, 

If you're on Java 11 and happy to compile a 1.3.1-SNAPSHOT build, I've made a [pull request](https://github.com/neo4j-graphql/neo4j-graphql-java/pull/229) for an initial implementation of date/time filtering. 
As @Andy2003 has pointed out, the `_Neo4jDateInput` type et al will be refactored out as part of the API alignment effort with `neo4j-graphql-js` - so that PR is highly unlikely to be merged.

If you're on Java 8, I've got a variant of that PR that I'm trialling, but it's on a branch created from the 25th March 2021 commit 0f89393884914e906de9d78c1e432b5a09a2ce53 before the update to Neo4j 4.2. I've not yet pushed that to my fork though.

Issue #93, opened 2020-07-10
Filter with Datetime (Temporal) not working


Techyoga1990:
I am trying to filter data using date time but Neo4J translator is not creating correct query, i am not sure if this is bug or feature is missing. 

I have schema like below :- 

```
type Person{
name: String
born: _Neo4jDateTime
} 
```

i am using below query using filter :- 

```
{
person(filter: {born: {year_gt: 2010} }){
name
born{
formatted
}
}
}
```

below Cyper is getting created :- 

```
MATCH(person:Person) where person.born > $filterPersonBorn_GT
RETURN ....

```
year is not getting considered in query. 

Can you please help me right query for date time filter or if there any workaround for this issue.

Andy2003:
This type of filtering for temporal types is not yet implemented and there is currently no workaround to get it working.
I will have a look at this issue.

jexp:
could we implement this by using the accessory like `born.year > $x `

Andy2003:
as pointed out in https://github.com/neo4j-graphql/neo4j-graphql-java/pull/229#issuecomment-845288588, the way to filter dates / times should be refactored to macht the logic of the js version.

This means, we remove the date-/time-related types and use scalars (iso strings) instead

PR #229, opened 2021-05-20
Add datetime comparisons for filtering (#93)

rbramley:
Issue #93 (plus others like #223) requires the ability to perform filtering with simple date comparisons.

This patch determines whether a type is a Neo4j Date/Time and augments it with filters for GT/GTE/LT/LTE (over and above the existing augmentation for a Neo4j type).


Andy2003:
Currently we are trying to harmonize our API with that of the official JavaScipt version (https://github.com/neo4j/graphql).
In this sense we should replace the (input-) types like `_Neo4jDateTime` etc. by simple scalars which transfer the date / time as iso string.
With this in place you would pass a complete date / time iso string in the filter.

see also:

https://github.com/neo4j/graphql/blob/master/packages/graphql/tests/tck/tck-test-files/schema/types/datetime.md
https://github.com/neo4j/graphql/blob/master/packages/graphql/tests/tck/tck-test-files/cypher/types/datetime.md

This means we need some refactoring:

* configuration to switch behavior until next major release
* implementation of the new scalars
* adjusting of the generated cypher

Andy2003:
The adjustments mentioned in https://github.com/neo4j-graphql/neo4j-graphql-java/pull/229#issuecomment-845288588 are implemented via  #235

Issue #87, opened 2020-06-12
nested null handling / propagation


jexp:
this fails

```
        // errors on non-null fields are propagated up to the first nullable field
        // and an error message is added
        assertResult("""{ p: person { name, company2(name: "ACME2") { name } } }""",
                mapOf("p" to listOf(null)))


    @Test
    fun dynamicMutationConflictsWithAutoGenerated() {
        val schema = """
type Person {
    name: String
    company: Company @relation(name:"WORKS_AT")
    company2: Company! @relation(name:"WORKS_AT")
    companies: [Company]! @relation(name:"WORKS_AT")
    companies2: [Company!]! @relation(name:"WORKS_AT")
}
type Company {
    name: String
}
"""
```

Andy2003:
The example above is taken from https://github.com/neo4j-graphql/neo4j-graphql/blame/3.5/src/test/java/org/neo4j/graphql/NestedNonNullTest.kt

It was introduced by https://github.com/neo4j-graphql/neo4j-graphql/issues/80

This feature requires post processing in combination with e.g. `apoc.util.validate`

As of now solving this issue has no priority.

PR #270, opened 2022-03-25
Resolve field alias in Cypher only if no DataFetchingInterceptor is used

### LINKED ISSUE #267: Field aliasing does not work if using DataFetchingInterceptor
**Describe the bug**

If using `DataFetchingInterceptor` field aliasing does not work, `null` is always returned and it does not matter whether the aliased type was a scalar or a complex type.

**Test Case**

***GraphQL schema***

```
type Movie {
  title: String
}
type Query {
  movie: [Movie]
}
```

***GraphQL request***

```
query {
  movie {
    someAlias: title
  }
}
```

***Expected cypher query***

since `graphql-java` is handling aliasing the cypher should look like this:

```
MATCH (movie:Movie)
RETURN movie {
	.title
} AS movie
```

instead it looks like this:

```
MATCH (movie:Movie)
RETURN movie {
	someAlias: movie.title
} AS movie
```

***Expected cypher params***

```
```

***Neo4j test data***

```
```

***Expected GraphQL response***

```
{
	"errors": [],
	"data": {
		"movie": [
			{
				"someAlias": "value representing the title"
			}
		]
	},
	"extensions": null,
	"dataPresent": true
}
```

**Additional context**

If using `Translator` this works fine because you get the cypher already with alias and you just run it and return the data as it was fetched by the cypher. If using `DataFetchingInterceptor` the cypher still contains the alias so the fetched data looks god at first sight, but when returned from the `DataFetchingInterceptor` `graphql-java` does a mapping using the type definition to return with the information about aliasing used in the request query. So it completely ignores the `someAlias` data in the result because it is not part of the type definition and instead it looks for `title` which it expects in the result (since it is not there `null` is returned) to later alias it with `someAlias` by itself.

Note that aliasing on query level works fine because you just fetch the data from neo4j using the `variable` (which can be an alias) but then return just the data from the `DataFetchingInterceptor` without the `variable` (without the alias) and aliasing is handled by `graphql-java` again and this time correctly of course.

So while fixing that take into account that the `Translator` approach might get broken.

Andy2003:
resolves #267

### DIFF
--- a/.run/Run All tests including integration tests.run.xml
+++ b/.run/Run All tests including integration tests.run.xml
@@ -1,6 +1,6 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="Run All tests including integration tests" type="JUnit" factoryName="JUnit">
-    <module name="neo4j-graphql-java (1)" />
+    <module name="neo4j-graphql-java"/>
     <option name="PACKAGE_NAME" value="org.neo4j.graphql" />
     <option name="MAIN_CLASS_NAME" value="" />
     <option name="METHOD_NAME" value="" />
--- a/core/src/main/kotlin/org/neo4j/graphql/GraphQLExtensions.kt
+++ b/core/src/main/kotlin/org/neo4j/graphql/GraphQLExtensions.kt
@@ -108,10 +108,7 @@ fun GraphQLType.ref(): GraphQLType = when (this) {
     else -> GraphQLTypeReference(name())
 }
 
-fun Field.aliasOrName(): String = (this.alias ?: this.name)
-fun SelectedField.aliasOrName(): String = (this.alias ?: this.name)
-fun SelectedField.contextualize(variable: String) = variable + this.aliasOrName().capitalize()
-fun SelectedField.contextualize(variable: SymbolicName) = variable.value + this.aliasOrName().capitalize()
+fun SelectedField.contextualize(variable: SymbolicName) = variable.value + (this.alias ?: this.name).capitalize()
 
 fun GraphQLType.innerName(): String = inner().name()
         ?: throw IllegalStateException("inner name cannot be retrieved for " + this.javaClass)
--- a/core/src/main/kotlin/org/neo4j/graphql/GraphQLExtensions.kt
+++ b/core/src/main/kotlin/org/neo4j/graphql/GraphQLExtensions.kt
@@ -108,7 +108,10 @@ fun GraphQLType.ref(): GraphQLType = when (this) {
     else -> GraphQLTypeReference(name())
 }
 
-fun SelectedField.contextualize(variable: SymbolicName) = variable.value + (this.alias ?: this.name).capitalize()
+fun Field.aliasOrName(): String = (this.alias ?: this.name)
+fun SelectedField.aliasOrName(): String = (this.alias ?: this.name)
+fun SelectedField.contextualize(variable: String) = variable + this.aliasOrName().capitalize()
+fun SelectedField.contextualize(variable: SymbolicName) = variable.value + this.aliasOrName().capitalize()
 
 fun GraphQLType.innerName(): String = inner().name()
         ?: throw IllegalStateException("inner name cannot be retrieved for " + this.javaClass)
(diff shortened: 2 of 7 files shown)


Andy2003:
@Krejtcha can you please check this PR if it solves your issue correctly?

Krejtcha:
@Andy2003 thank you for the fix, but unfortunately it fixes the problem only partially. Now aliases on fields work, but the main purpose of why aliases are used does not work. I did not have time to investigate the issues in detail so I do not know the cause. I can only provide you with this info:

- if you query for the same neo4j property multiple times and present a different alias it works fine: e.g. `aaa: identifier`, `bbb: identifier`
- if you query for the same "special" property like `_id` or ap property driven by `@cypher` directive e.g. `aaa: _id`, `bbb: _id` the library does not even come to the part where it tries to fetch data from neo4j and you will get a `DataFetchingException`: `Exception while fetching data (/someNode) : Duplicate key '_id'`
- the same happens if you query for the same relationship multiple times e.g. `aaa: RELATIONSHIP_TO_OtherNode (<some filter>) ...`, `bbb: RELATIONSHIP_TO_OtherNode (<some other filter>) ...` you will get a `DataFetchingException`: `Exception while fetching data (/someNode) : Duplicate key 'RELATIONSHIP_TO_OtherNode'` - this is the use case why aliases are needed because you can not do that without them - so this needs to work.

Hopefully this feedback will be helpful for you and you will be able to fix this issue.

Andy2003:
@Krejtcha thank you for your feedback! I now had to implement a completely different approach. Please check again if the PR fixes your issue correctly now.

Krejtcha:
@Andy2003 unfortunately is still does not work:

If there is a `@cypher` based property e.g. `_label: String!` defined as not null and I query:

```
	other {
                _label
		l: _label
	}
```

it is OK, but when I query just this:

```
	other {
		l: _label
	}
```

I get `NullValueInNonNullableField`: `The field at path '/other[0]/l' was declared as a non null type, but the code involved in retrieving data has wrongly returned a null value.  The graphql specification requires that the parent field be set to null, or if that is non nullable that it bubble up null to its parent and so on. The non-nullable type is 'String' within parent type 'Other'`

It behaves the same for `@relation` based properties and will probably behave the same for standard properties as well.

If I have a rich relationship and I identify the `from` node as `_start` and `to` node as `_end` and query the following:

```
	other {
		Test_RELATED_TO {
			_start {
				identifier
			}
			s: _start {
				identifier
			}
		}
	}
```

it works, but when query just this:

```
	other {
		Test_RELATED_TO {
			s: _start {
				identifier
			}
		}
	}
```

`s` is null.

If I query this:

```
	other {
		Test_RELATED_TO (_id: 48389) {
			_id
			_start {
				identifier
			}
		}
		r1: Test_RELATED_TO (_id: 0) {
			_id
			_start {
				identifier
			}
		}
	}
```

both `_id` for `r1` and for not aliased are the same: `48389` and if I query this:

```
	other {
		r1: Test_RELATED_TO (_id: 0) {
			_id
			_start {
				identifier
			}
		}
	}
```

again the error with `NullValueInNonNullableField` sicnce `Test_RELATED_TO` is defined as `[Test_RELATED_TO_Other!]!`

Andy2003:
@Krejtcha I cant reproduce your described behavior, can you do me a favor and adjust https://github.com/neo4j-graphql/neo4j-graphql-java/pull/270/files#diff-39e36965c8869d32dea7d00ef257118ec05f65fac67eb0a4af6fd9e926de64b0 to represent your descibed behavior.

Could it be that you are stumbling over another issue: https://github.com/neo4j-graphql/neo4j-graphql-java/issues/87?

Andy2003:
I moved the logic to `registerDataFetcher`, so your approach should work now.
In the examples we do the same as you:

https://github.com/neo4j-graphql/neo4j-graphql-java/blob/689ded127c117bd559f425436be05cfbcc6616c7/examples/dgs-spring-boot/src/main/kotlin/org/neo4j/graphql/examples/dgsspringboot/config/GraphQLConfiguration.kt#L52-L56

We could add some customizer callbacks (a.k.a. extension points) as a new feature. Feel free to create a feature request. PR's are of course always welcome as well!

PR #276, opened 2022-06-14
update graphql-java dependency to 18.1

caochun:
update graphql-java dependency to 18.1 ( 17.3 is actually not compatible )


Andy2003:
resolved via #279

PR #296, opened 2023-08-24
updated logic for endNodePattern to include label information in the variable name

### LINKED ISSUE #295: Wrong node name used for end node in rich relationship
**Description**
In our schema we have defined rich relationship. When we are trying to filter on the end node of the relationship, it is throwing error.

***GraphQL schema***

```
type Person{
    name: String
    age: Int
    rel_has_target_book: [Person_HAS_Target_Book]
}

type Book{
    name: String
    price: Int
}

type Person_HAS_Target_Book @relation(name: "HAS", from: "source", to: "target", direction: OUT) {
    name: String
    source: Person
    target: Book
}
```

***GraphQL request***

```
query{
  person{
    name
    rel_has_target_book(where:{target:{name:"Book 1"}}){
      target{
        name
      }
    }
  }
}
```

***Generated Cypher query***
Here, "personRel_has_target_bookTarget" is the variable name for Book but in the where condition, "personRel_has_target_bookBook" is being used
```
MATCH (person:Person)
CALL {
    WITH person
    MATCH (person)-[personRel_has_target_book:HAS]->(personRel_has_target_bookTarget:Book)
    WHERE personRel_has_target_bookBook.name = $personRel_has_target_bookBookName
    RETURN collect(personRel_has_target_book {
        target: personRel_has_target_bookTarget {
            .name
        }
    }) AS personRel_has_target_book
}
RETURN person {
    .name,
    rel_has_target_book: personRel_has_target_book
} AS person

Variables: {personRel_has_target_bookBookName=Book 1}
```

***Expected cypher query***

```
MATCH (person:Person)
CALL {
    WITH person
    MATCH (person)-[personRel_has_target_book:HAS]->(personRel_has_target_bookBook:Book)
    WHERE personRel_has_target_bookBook.name = $personRel_has_target_bookBookName
    RETURN collect(personRel_has_target_book {
        target: personRel_has_target_bookBook {
            .name
        }
    }) AS personRel_has_target_book
}
RETURN person {
    .name,
    rel_has_target_book: personRel_has_target_book
} AS person

Variables: {personRel_has_target_bookBookName=Book 1}
```

**Additional context**
We looked into the codebase. 
In the file "**ProjectionBase.kt**",  if we modify the below line in the function "**projectRichAndRegularRelationship**", then the correct cypher query is being generated.

Before: node(label).named("$childVariable${**relInfo.endField**.capitalize()}") to **relInfo.endField**

After: node(label).named("$childVariable${**label**.capitalize()}") to **label**

saurav213:
updated logic for endNodePattern to include label information in the variable name.

It fixes #295

### DIFF
--- a/core/src/test/resources/filter-tests.adoc
+++ b/core/src/test/resources/filter-tests.adoc
@@ -3611,6 +3611,20 @@ RETURN person {
 } AS person
 ----
 
+.GraphQL-Response
+[source,json,response=true,ignore-order]
+----
+{
+  "person" : [ {
+    "cards" : [ {
+      "target": {
+        "name": "Mastercard"
+      }
+    }]
+  } ]
+}
+----
+
 '''
 
 === Cascaded filter object string _eq_
--- a/core/src/test/resources/filter-tests.adoc
+++ b/core/src/test/resources/filter-tests.adoc
@@ -90,20 +90,7 @@ SET p = props;
 
 CREATE (c:Card {name: 'Mastercard'})
 WITH c
-UNWIND [{
-  id:       'jame',
-  name:     'Jame',
-  age:      35,
-  hobbies:  ["Cycling", "Singing"],
-  gender:   'male',
-  fun:      true,
-  height:   1.64,
-  location: point({longitude: 3, latitude: 4, height: 5})
-}]
-AS props
-
-CREATE (p:Person)-[:HAS_CARD]->(c)
-SET p = props;
+CREATE (p:Person)-[:HAS_CARD]->(c);
 ----
 
 == Filter on _List_
--- a/core/src/test/resources/filter-tests.adoc
+++ b/core/src/test/resources/filter-tests.adoc
@@ -18,6 +18,7 @@ type Person {
   company: Company @relation(name:"WORKS_AT")
   skills: [PersonSkills]
   location: _Neo4jPoint
+  cards: [Cards]
 }
 type Company {
   _id: ID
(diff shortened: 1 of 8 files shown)


Andy2003:
Can you add a test that coveres the bug  (with the wrong where-parameters)

saurav213:
Added a test case to validate variable name in where condition.

core/src/test/resources/filter-tests.adoc
```diff
@@ -72,7 +86,24 @@ UNWIND [{
 AS props
 
 CREATE (p:Person)-[:WORKS_AT]->(c)
-SET p = props
+SET p = props;
```
Andy2003, line 89: why adding test data, if you don't use it in the tests?

saurav213: In below tests, there are graphql queries which are filtering on node properties. So this "props" is required here.

But in my test case for "wrong variable name in where condition", props is not required. I will remove it from there.

saurav213: Removed unwanted data from my test case

Andy2003: What I meant is that if you add test data, which is perfectly OK, then the test should also check if a particular query returns that data. This is tested using the following block:

```adoc
.GraphQL-Response
[source,json,response=true,ignore-order]
----
{
 "expected": "graphql response"
}
----
```

Andy2003 (commented):
Do you want to add integration test assertions?

saurav213:
Hi.
I will add it today. I got held up with some other work.

core/src/test/resources/filter-tests.adoc
saurav213: Added Graphql Response

saurav213:
> Do you want to add integration test assertions?


You are talking about adding this response in my test, right?

```
.GraphQL-Response
[source,json,response=true,ignore-order]
----
{
 "expected": "graphql response"
}
----
```

Andy2003:
Please fix the tests!

saurav213:
> Please fix the tests!

In my local, all the CypherTests are running fine.

picc

After running mvn clean install, I don't see any test failure in Neo4j GraphQL Java 1.7.1-SNAPSHOT 

image

Andy2003:
I will take a look at this

PR #310, opened 2024-01-31
Support augmentation of enum fields

isstabb:


isstabb:
@Andy2003 What do you think? I wasn't sure if this was a bug so I did not create one.

Andy2003:
@isstabb thank you for your contribution!

